PR #253, opened 2013-07-18
Prefetch Loading. This adds prefetch support. mouseover and touchstart events are tracked shaving off up to a few hundred ms in response reaction.

davydotcom:
This prefetch support will monitor touchstart or mouseover events. If detected an ajax fetch will be called and cached. When the click event occurs if the cache is finished loading it will use it. If not, it will set the state such that the xhr request will process the load as a standard load. Shaving off request delay cycles.


Thibaut:
Closing since no activity in more than a year. Feel free to open a new PR on the latest `master`.

PR #291, opened 2013-10-22
Add possibility to keep current scroll position in the Turbolinks.visit API

tzvetkoff:
On certain occasions some might need to keep current scroll position.
In my case I call `Turbolinks.visit(location.href)` to reload the content of the current page -- this is an improvised replacement for `location.reload()`.
Problem is, that if you want to reload the current page content for, say, new elements created by AJAX form submission, the page will scroll back to top, yet you want to keep the scroll position.

Partial replacements is a way to do this, but requires specific Rails code on the backend to just render that partial and then you're getting even more tied to the front-end.

So, I've added an extra option to keep the scroll position.

When invoking Turbolinks manually:
```javascript
Turbolinks.visit(path, { keepScrollPosition: true });
```

Or by marking the element with the `data-turbolinks-keep-scroll` attribute:
```html
<a href="/path" data-turbolinks-keep-scroll>Change the page and keep current scroll position.</a>
```

Edited: **2015-05-04**

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,22 @@ You can use `Turbolinks.visit(path)` to go to a URL through Turbolinks.
 
 You can also use `redirect_to path, turbolinks: true` in Rails to perform a redirect via Turbolinks.
 
+Scroll position (3.0+)
+----------------------
+
+By default, `Turbolinks.visit(path)` will reset the scroll position to `(0, 0)`.
+
+You can prevent that by invoking it like `Turbolinks.visit(path, { keepScrollPosition: true })`, or by marking your link or any parent container with the `data-turbolinks-keep-scroll` attribute.
+
+```html
+<a href="/path" data-turbolinks-keep-scroll>Refresh the page and keep the scroll position.</a>
+
+<script type="text/javascript">
+  // Call it manually
+  Turbolinks.visit('/path', { keepScrollPosition: true });
+</script>
+```
+
 Partial replacements with Turbolinks (3.0+)
 -------------------------------------------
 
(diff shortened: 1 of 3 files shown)


drewhamlett:
I believe this will be more important with the new Turbolinks 3 partial changes.  Right now with partial changes the page scrolls to the top which is not what you really want happening.  

Right now I'm doing this.  Which was posted in one of the other issues about scrolling.

```coffee
$(document).on "page:change", ->
  window.prevPageYOffset = window.pageYOffset
  window.prevPageXOffset = window.pageXOffset

$(document).on "page:load", ->
  window.scrollTo window.prevPageXOffset, window.prevPageYOffset
```

Thibaut:
> Right now when the partial changes the page scrolls to the top.

@drewhamlett This shouldn't happen with `Turbolinks.replace` / `render :action, change: :key`. Can you post the code that you're using?

drewhamlett:
I must be missing something, because I'm using the redirect to in this case.  I have a section on a page that gets updated, but the page requires a bunch of other data to re render that page, not specific the section that gets updated.  If I just use render I have to duplicate a lot of code to fetch all the other pieces of the page.

Unless I'm doing something wrong?

Thibaut:
@drewhamlett Oops, I forgot that `Turbolinks.visit` / `redirect_to` can also do partial replacements. Expect a PR / fix soon.

tzvetkoff:
Rewrote the patch to conform to the latest Turbolinks code and updated the PR description.

Thibaut:
@tzvetkoff I'm working on a similar PR with tests for some of the edge cases and a fix for the problem @drewhamlett described. Will keep you updated.

vizo:
:+1:  Thanks guys, very useful. Page refreshing will look like on SPA :)

Thibaut:
`Turbolinks.visit(url, scroll: false)` has been merged into master (#528). I will PR `data-turbolinks-scroll=false` and update the docs this week-end.

Thanks all for pushing this forward.

ajb:
Thanks @Thibaut and everyone!

PR #300, opened 2013-11-19
Integrated header check for redirects via redirect_with_turbolinks to avoid plain body responses on non AJAX requests

blinkist:
As we got some troubles with *redirect_via_turbolinks* and plain text responses in some situations, I investigated on the problem and found this short snippet of code that doesn't distinguish between regular http requests and those sent via XMLHttpRequest. 

With my fix, we never experienced a 200 response with *Turbolinks.visit(...)* in the body anymore. All regular non-AJAX requests got redirected perfectly by the browser. AJAX requests get the javascript response and execute the turbo links code.

### DIFF
--- a/lib/turbolinks.rb
+++ b/lib/turbolinks.rb
@@ -53,9 +53,13 @@ module Turbolinks
     def redirect_via_turbolinks_to(url = {}, response_status = {})
       redirect_to(url, response_status)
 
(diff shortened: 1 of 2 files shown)


reed:
This isn't an issue anymore, since a check if the request is XHR has been added to the new redirect code.

PR #331, opened 2014-01-21
Prevent updating document url to the same url

mattsjohnston:
Currently when using pushState in an app, it's possible that turbolinks will update the document url to an identical url, which simply causes an unnecessary refresh of the current page.

This update checks to make sure it's not already set to the correct url before trying to update it.


reed:
Closing this since the proposed change would break the fallback behavior when the response isn't valid.

PR #356, opened 2014-03-29
Mechanism to override body replace

mrloop:
This commit keeps the default behaviour but lets you override the default behaviour of replacing the body tag.

Here's an example of me overriding the default behaviour.
```coffee  
$ ->
  $(document).on 'page:replace', (e) ->
    turboElement = document.getElementById 'turbolinks-replace'
    newTurboElement = $(e.originalEvent.data.html).find('#turbolinks-replace')[0]
    turboElement.parentElement.replaceChild newTurboElement, turboElement  
    false #cancel default turbo link replace
```

I've got a long running video background I don't want reloaded every page replace.

https://github.com/rails/turbolinks/pull/204, I also ran into the page caching issue described here when overriding the default behaviour

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -31,7 +31,7 @@ transitionCacheFor = (url) ->
 enableTransitionCache = (enable = true) ->
   transitionCacheEnabled = enable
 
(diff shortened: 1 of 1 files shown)


reed:
Closing in favor of the partial page replacement feature introduced in v3.0.  Should give you the functionality you need.

Issue #66, opened 2012-10-02
reset scroll position on page replacement


moomerman:
It's really annoying then you visit a link with shorter content and being greeted with a blank page, having to scroll up to see the content.  I just experimented with resetting the scroll position to 0,0 on successful fetch and it feels much more natural:

```
xhr.onload  = ->
    changePage extractTitleAndContent(xhr.responseText)...
    reflectRedirectedUrl xhr
    resetScrollPosition()
    triggerEvent 'page:load'

resetScrollPosition = ->
  window.scrollTo 0, 0
```

PR #359, opened 2014-04-07
Option to keep scroll position on page change

linesarefuzzy:
This patch enables attaching a “data-keep-scroll” attribute to a link, and clicking it will keep the window’s scroll position instead of scrolling to the top of the page. 

This is a response to this issue: https://github.com/rails/turbolinks/issues/66#issuecomment-14853274
I needed more fine grained control, with some links keeping scroll position and others expressing the default behavior. It's working great for me with this patch. Hopefully this will be helpful to others as well.


Thibaut:
Fixed by #528.

PR #372, opened 2014-05-02
Added rememberCurrentState and cacheCurrentPage to the Public API

jjoos:
We needed some custom behaviour in our application, we want to cache dynamic pages (backbone/marionette).

We always call these two methods `rememberCurrentState` and `cacheCurrentPage` together, so creating one method that does both would also work for us.

If we completely missed a better way of doing this, please let us know!


Thibaut:
I don't understand why you're calling `pushState` / `replaceState` manually and would prefer not exposing Turbolinks's internals like that.

Feel free to reopen if this is still an issue for you.

PR #378, opened 2014-05-14
Added a before-fetch event

rldaggie:
Hey guys,

I added a ```page:before-fetch``` event to intercept the back button for an app I'm working on. I'm brand new to contributing to OS, apologies if I'm not following protocol.


reed:
Closing due to the issues I previously outlined, the fact that browsers don't provide a way to cancel the back button, and the lack of response from submitter.

PR #379, opened 2014-05-19
[WIP] Add form/POST support

Thibaut:
At Shopify we've extended Turbolinks to support partial DOM replacement in order to submit all our forms and links via AJAX and — instead of using server-generated JavaScript responses — update portions of the page without losing focus and other DOM state.

Our current implementation uses jQuery to submit forms and only passes the successful `xhr` object back to Turbolinks to refresh the page. Ideally this should be fully automatic in Rails/Turbolinks but this requires adding POST/form support to Turbolinks. I'm not entirely sure this is the best approach since it brings "hard to maintain" code to Turbolinks, but we wanted to give it a shot and open up the discussion (before extracting partial page replacement from Shopify).

The main challenges are form serialization, which is harder to do and maybe less reliable with plain JavaScript than `jQuery.fn.serialize`, and what to do when the server responds with a response that Turbolinks can't handle (e.g. 500). In that case I don't think Turbolinks should re-submit the form without AJAX as it does with links (maybe we could simply fire an event?).

Note that this code is just a proof of concept / untested. If we decide to go with this approach I'd be happy to write a full test suite / get it ready for production.

Another option might be to tie this in `jquery-ujs`.

@dhh @josh @arthurnn @jduff what do you think?

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -10,6 +10,7 @@ referer                 = null
 createDocument          = null
 xhr                     = null
 
+r20                     = /%20/g
 
 fetch = (url) ->
   url = new ComponentUrl url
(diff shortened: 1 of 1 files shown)


Thibaut:
Closing this. Shopify's version of Turbolinks (with the features described above) is available here:
https://github.com/Shopify/turbograft

dhh:
This looks pretty promising. I’d love to see some use cases for some of the more exotic use cases. Been thinking about introducing the simple “data-turbolinks-keep” to prevent certain elements from being refreshed on a new page. But of course graft has a bunch more stuff. Not sure I’m loving having full tg-remote stuff. That’s an awful lot of logic to burden the templates with imo. But let’s see if we can’t push some of the partial element replace stuff into Turbolinks.

PR #386, opened 2014-06-09
Do a document.importNode on the new body.

cwninja:
Setting the document.domain on the consuming page caused IE10 to error out with a WrongDocumentError.


Thibaut:
Tests pass on IE 10. Feel free to re-open if this is still an issue with the latest `master`.

PR #397, opened 2014-08-07
Fix back button + hash urls + page cache

masonjm:
With this sequence:

page1.html -> page1.html#foo -> page2.html -> (back button) -> (back button) -> page1.html#foo -> page2.html -> (back button)

I end up seeing the contents of page1.html instead of page1.html#foo. This was biting me on a page that uses javascript to show a different element based on the URL anchor.


Thibaut:
Feel free to re-open a PR on the latest `master` with a test case.

Issue #413, opened 2014-10-03
Turbolinks breaks resolution independent images on a large number of browsers


tomafro:
To recreate, add the following markup to a view that will be loaded via turbolinks:

```
<img src="example.png" srcset="example@2x.png 2x, example@3x.png 3x, example@4x.png 4x">
```

Using Chrome, open the Developer Tools, click the device emulation button and choose a 
device with a dots per pixel ration of greater than 1 (such as the iPhone 5). In the emulator
navigate to the view with the img tag (N.B. don't go directly to the page; turbolinks must
load the page for the bug to be triggered)

![Chrome device emulator](https://dl.dropboxusercontent.com/s/nvj59igw5xtrlnf/2014-10-03%20at%2010.29%202x%20%282%29.png?dl=0)

**What should happen:**

One of the high resolution images such `example@2x.png` or `example@3x.png` should be displayed
depending on the device.

**What does happen:**

`example.png` will be shown. If you refresh the page (loading it directly) the correct image
will be shown. This affects not only Chrome, but AFAIK all Webkit/Blink based browsers, including
crucially Safari on iOS and Chrome on Android. It doesn't affect Firefox nightly.

**Why does this happen?:**

The document factories `createDocumentUsingParser`, `createDocumentUsingDOM`, and 
`createDocumentUsingWrite` create new detached DocumentElements which until they're displayed
have an internal device pixel ratio (dpr) of 1. Image srcset selection takes place as the
document is built, not when it is displayed, so the lowest resolution image will always
be shown.

Possible solutions:

While this might arguably be a bug in Webkit, we'd like to use resolution independent images
now and so would appreciate a solution from turbolinks. Using the following create document 
function fixes this issue for me on Chrome and Safari on OSX, and on Android and iOS.  I believe 
it works because the clone of the documentElement has the same dpr as the existing document.

```
createDocumentWithClone = (html) ->
  doc = document.documentElement.cloneNode()
  doc.innerHTML = html
  doc.head = doc.querySelector 'head'
  doc.body = doc.querySelector 'body'
  doc
```

However, I don't know the full background behind the other implementations - whether they exist for efficiency or to work around other bugs - so I don't know if this introduces other problems. I also don't get how to run/use the existing tests.

I'd love to help fix this. Any guidance appreciated.

P.S. For easy testing, here are the images I've been using. They make it easy to see which resolution
[...]

PR #425, opened 2014-11-03
Use document.documentElement.cloneNode() when parsing to fix resolution independence

tomafro:
This follows on from issue https://github.com/rails/turbolinks/issues/413 (which has a fair bit more background info).

In order to fix the resolution independence issues we've been seeing (where the @1x image is always chosen even when higher resolutions would be appropriate) we've been using the `createDocument` function in this pull request, using document.documentElement.cloneNode() in place of the other document creation techniques. We've been using this on a staging environment and so far have seen no issues. 

For this initial pull request I've ripped out all of the other document creation functions as it seems cleaner to me to use a single function across all browsers with a fallback only where necessary. Unfortunately I'm not really clear how to check where a fallback would be necessary. I've done sanity checks on Chrome, Safari, Opera, IE10, iOS Safari, Android Chrome, and Firefox and it all looks good, but what else should I look at to add confidence?

I have written a performance check and it seems that using `cloneNode()` is generally faster (though performance is so high already there's no appreciable benefit): http://jsperf.com/compare-turbolinks-document-creation

Any comments, questions, and help getting this pull request either improved or rejected will be gratefully received.

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -8,7 +8,6 @@ loadedAssets            = null
 
 referer                 = null
 
-createDocument          = null
 xhr                     = null
 
 EVENTS =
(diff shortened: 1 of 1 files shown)


ybart:
Because of a customer using a very old version of Firefox (11.0) which required a parameter to cloneNode. The problem caused the links not working at all.

```
  doc = document.documentElement.cloneNode(true)
```

PR #449, opened 2015-01-06
Progress bar improvements

MGotink:


javan:
Thanks!

In my testing, I found that mobile Safari on the iOS simulator would crash when click a link a few times before the progress bar finishes. I traced it back to the `@element.offsetHeight` call -- commenting that out prevents the crash, but also prevents the necessary reflow. I couldn't reproduce the issue on a real device so I'm not sure it's worth investigating any further.

mobile-safari-crash

MGotink:
I've reproduced it once with the iOS simulator, but can't reproduce it anymore now. If it's not reproducible on a real device I wouldn't spend any more time on it, it could as well be an issue with the simulator.

Issue #448, opened 2015-01-06
Partial replacement of new pages


dhh:
Some great ideas from https://github.com/Shopify/turbograft that we should migrate into Turbolinks. The first is partial replacement. I was initially skeptical to this idea, but have been won over by seeing how it can really work well with permanent elements of a page like shopping carts or sidebars or other pieces with state. 

The big win from Turbolinks remain that the initial hit will generate the full page, and that you won't have to tailor server-side rendering on subsequent requests, even when you're only doing partial updates.

I'd like to propose the following API:

```
<div data-turbolinks-partial="section-a">
</div>

<div data-turbolinks-permanent>
</div>

<a href="#" data-turbolinks-change="section-a"></a>

Turbolinks.visit(url, change: [ 'section-a', 'section-b' ])
Turbolinks.visit(url, keep: [ 'section-c', 'section-d' ])
```

So you can name a partial with data-turbolinks-partial. These named partials can then be referred to in the Turbolinks.visit call through the change/keep options.

Marking an element with "data-turbolinks-permanent" will mean that it is not updated by any Turbolinks.visit call, except if you ALSO give the same element a partial name AND name that partial as part of a Turbolinks.visit/change call.

kristianpd:
@dhh interesting to see this discussion taking place. A few comments:


## keep and change vs. refresh-always, refresh-never and refresh

`data-turbolinks-permanent` seems equivalent to our [refresh-never](https://github.com/Shopify/turbograft/blob/master/lib/assets/javascripts/turbograft/turbolinks.coffee#L139) key.  We only use this twice in our code base to prevent a `script` tag from executing inline more than once (tracking script). We never really ended up using it as much as we had thought we would. I'm curious what use cases you may have for putting it on an element like a `div`. I also intended for it to be used this way, but we never did.  

[...]

PR #454, opened 2015-02-03
Add cacheCurrentPage to public API 

ilucin:
I think it would be good to add cacheCurrentPage to the public API of Turbolinks. I'm developing a complex rails application and got to a case where I desperately need it. I'll explain my case in details, but the whole idea can be applied generally.

**My case**
Web application that has many views that can be rendered as a normal page visit (on different URL) and inside of a modal dialog (URL stays the same). When a page is opened in a modal dialog Turbolinks.visit() doesn't get called, but the URL gets manually updated (using history.pushState) so it has the page route that was rendered in the modal. When the modal gets closed, history.back() is called to go to a previous (original) route. So this is about manipulating URL history stack when opening and closing a modal dialog.

**My problem**
Some pages are complex and they can have dynamic HTML updates (drag'n'drops, adding & removing elements with ajax ...) and Turbolinks doesn't properly cache those changes. So when some changes are made on a page, then opened another page in the modal dialog (pushState) and then navigated back to the first page - Turbolinks shows wrong HTML (pulled from cache), without those changes made previously.

**My solution**
Added cacheCurrentPage() to public API of Turbolinks and calling it before each history.pushState(newUrl) fixed all issues.

**Good things if you merge**
* Nothing can break when cacheCurrentPage is exposed
* Added more flexibility to the library because now apps can push custom URL state without caching issues

**Bad things if you merge**
* Added more complexity to the public API and documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,13 +57,18 @@ Turbolinks.pagesCached();
 Turbolinks.pagesCached(20);
 ```
 
-When a page is removed from the cache due to the cache reaching its size limit, the `page:expire` event is triggered.  Listeners bound to this event can access the cached page object using `event.originalEvent.data`.  Keys of note for this page cache object include `url`, `body`, and `title`.  
+If you need to make dynamic HTML updates in the current page and want it to be cached properly you can call:
+```javascript
+Turbolinks.cacheCurrentPage();
+```
+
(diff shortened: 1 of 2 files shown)


dhh:
:+1: – if you're interested in helping us work to improve the precision of replacement in TL, please checkout https://github.com/rails/turbolinks/issues/448

PR #456, opened 2015-02-11
Use Jquery For Bindings

AlexRiedler:


AlexRiedler:
err, how did this happen... my bad

PR #459, opened 2015-02-14
cloudflare compatibility

fruwe:
We had an issue with turbolinks, which caused inline javascript not being executed, if the service is behind cloudflare with rocket loader enabled.

I think what happened is that turbolinks tried to fetch the next page, then cloudflare replaced all javascript script tags with rocketscript tags, which neither rocket.js nor turbolinks executed.

I added the ability to execute rocketscript to the executeScriptTags function.

Please, let me know, what you think.

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -180,9 +180,12 @@ swapNodes = (targetBody, existingNodes, options) ->
 
 executeScriptTags = ->
   scripts = document.body.querySelectorAll 'script:not([data-turbolinks-eval="false"])'
-  for script in scripts when script.type in ['', 'text/javascript']
+  for script in scripts when script.type in ['', 'text/javascript', 'text/rocketscript']
     copy = document.createElement 'script'
-    copy.setAttribute attr.name, attr.value for attr in script.attributes
+    copy.setAttribute attr.name, attr.value for attr in script.attributes when attr not in ['data-rocketsrc']
+    if script.type == 'text/rocketscript'
+      copy.setAttribute 'src', script.attributes['data-rocketsrc'].value if script.attributes['data-rocketsrc']
+      copy.setAttribute 'type', 'text/javascript'
     copy.async = false unless script.hasAttribute 'async'
     copy.appendChild document.createTextNode script.innerHTML
     { parentNode, nextSibling } = script


reed:
I think the more ideal solution would be to somehow get the rocket loader to run on `page:load`.  Have you looked into that?  Because I would imagine that the rocket.js loads the rocketscripts differently than turbolinks loads regular scripts, so it would be better if the rocketscripts were executed the way that they were intended to be.  

If it's not possible to do that, then I think that it'd be best if you handled the execution of the rocketscripts in your own code.  Essentially take what you put into this PR and run it on `page:load`.  Because if we add support for this third party feature, then we'd be opening ourselves up to having the obligation to support all the other third party libraries.

naveed-ahmad:
You can simply disable rocketscript for some script tags by adding `data-cfasync => false`

``` ruby
    javascript_tag 'data-cfasync' =>  false do
      your JS stuff here and rocketscript won't mess with it
    end
```

kartikluke:
I'm seeing the same issue @fruwe. How did you fix it?

fruwe:
@kartikluke Unfortunately I did not have time to figure out the best approach, furthermore disabling rocketscript is also not a viable option, I am still using my branch.
https://github.com/fruwe/turbolinks/tree/cloudflare_compatibility

kartikluke:
I have opened up a ticket and will discuss it with them. Your solution will work but as you can probably tell you lose all the benefits of RocketLoader by this method on pages loaded by Turbolinks.

fruwe:
Thanks a lot. The part I changed is only executed after a page transition via Ajax, when all external scripts are already loaded and only embedded scripts are executed. Thus you should be able to keep 99% of the benefits of RocketLoader, I think

kartikluke:
Yeah that's true. You lose the benefits for the embedded scripts. But that's not too bad. I think for now we should find a way to run your script outside of Turbolinks so we can keep using the latest turbolinks. @reed Could you suggest a way to do that?

fruwe:
I agree
(By the way, I just merged the latest turbolinks into my branch)

kartikluke:
Cool I think I'll use your branch for now. The RocketLoader made for quite a speed boost.

fruwe:
Sorry, for my late reply. I do not recall, but I think, I had some timing issues with things, I wanted to execute after a page has been loaded.

No, warranties, but if you run into such problems you might use this. (If somebody knows a better solution, please tell me. I am not familiar with jquery queues)

```haml
- flash.each do |name, msg|
  - content_for :page_bottom do
    :coffee
      document.queueAfterLoad () ->
        $.notify({message: '#{msg.to_s.html_safe}', type: '#{(name != :error ? 'danger' : 'info')}'})

...
  
= yield :page_bottom
```

```coffee
# queue of functions
document.afterLoadQueue = $({});
# queue processing started flag
document.afterLoadStarted = false

# queue or process immediately
document.queueAfterLoad = (func) ->
  if document.afterLoadStarted
    func()
  else
    document.afterLoadQueue.queue 'afterLoadQueue', (next) ->
      func()
      next()

# process queue
document.dequeueAfterLoad = () ->
  document.afterLoadStarted = true
  document.afterLoadQueue.dequeue 'afterLoadQueue'

$ ->
  # called once
  document.dequeueAfterLoad()
  $(document).on 'page:fetch', ->
    # called when page will change
    document.afterLoadStarted = false
  $(document).on 'page:load, page:restore', ->
    # called after page loaded
    document.dequeueAfterLoad()
```

fruwe:
this might interest you as well:
https://github.com/kossnocorp/jquery.turbolinks 

although, i have not tried it in conjunction with cloudflare, I tried this gem and it seems to be a "drop in, solve all my troubles and forget it" solution.

PR #468, opened 2015-02-27
Partial replacement 

### LINKED ISSUE #448: Partial replacement of new pages

kristianpd:
This closes #448 by adding partial replace capabilities to Turbolinks.

@Thibaut @dhh @arthurnn

### DIFF
--- a/lib/turbolinks/redirection.rb
+++ b/lib/turbolinks/redirection.rb
@@ -2,8 +2,6 @@ module Turbolinks
   # Provides a means of using Turbolinks to perform redirects.  The server
   # will respond with a JavaScript call to Turbolinks.visit(url).
   module Redirection
-    extend ActiveSupport::Concern
-
     def redirect_to(url = {}, response_status = {})
       turbolinks = response_status.delete(:turbolinks)
       options = response_status.extract!(:keep, :change, :flush)
--- a/lib/turbolinks/redirection.rb
+++ b/lib/turbolinks/redirection.rb
@@ -19,12 +19,14 @@ module Turbolinks
     def render(*args, &block)
       render_options = args.extract_options!
       turbolinks = render_options.delete(:turbolinks)
+      options = render_options.extract!(:keep, :change)
+      raise ArgumentError, "cannot combine :keep and :change options" if options.size > 1
 
       super(*args, render_options, &block)
 
-      if turbolinks
+      if turbolinks || (turbolinks != false && options.size > 0 && request.xhr? && !request.get?)
         self.status           = 200
-        self.response_body    = "Turbolinks.replace('#{view_context.j(response.body)}');"
+        self.response_body    = "Turbolinks.replace('#{view_context.j(response.body)}'#{_turbolinks_js_options(options)});"
         response.content_type = Mime::JS
       end
     end
--- a/lib/turbolinks/redirection.rb
+++ b/lib/turbolinks/redirection.rb
@@ -5,27 +5,15 @@ module Turbolinks
     extend ActiveSupport::Concern
 
     def redirect_to(url = {}, response_status = {})
-      change = response_status.delete(:change)
+      options = response_status.extract!(:change, :turbolinks)
       super(url, response_status)
 
-      if request.xhr? && !request.get?
-        perform_turbolinks_visit(change)
+      if options[:turbolinks] || (request.xhr? && !request.get?)
+        change = ", { change: ['#{Array(options[:change]).join("', '")}'] }" if options[:change]
+        self.status           = 200
+        self.response_body    = "Turbolinks.visit('#{location}'#{change});"
+        response.content_type = Mime::JS
       end
-    end
-
-    def redirect_via_turbolinks_to(url = {}, response_status = {})
-      change = response_status.delete(:change)
-      redirect_to(url, response_status)
-      perform_turbolinks_visit(change)
-    end
-
-    private
-
-    def perform_turbolinks_visit(change = nil)
-      change = ", { change: ['#{Array(change).join("', '")}'] }" if change
-      self.status           = 200
-      self.response_body    = "Turbolinks.visit('#{location}'#{change});"
-      response.content_type = Mime::JS
     end
   end
 end
--- a/lib/turbolinks/redirection.rb
+++ b/lib/turbolinks/redirection.rb
@@ -6,8 +6,8 @@ module Turbolinks
 
     def redirect_to(url = {}, response_status = {})
       turbolinks = response_status.delete(:turbolinks)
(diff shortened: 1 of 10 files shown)


Thibaut:
**tldr;**

```html
<body>
  <div id="sidebar" data-turbolinks-permanent>
    Sidebar, never changed!
  </div>
 
  <div id="flash" data-turbolinks-temporary>
    You have new comments. <%= link_to 'See new comments', comments_path, change: 'comments' %>
  </div>
 
  <section id="comments:collection">
    <article id="comment_123"></article>
  </section>
 
  <div>
    <a id="comments:link">See 45 more comments</a>
  </div>
 
  <%= form_for Comment.new, remote: true, id: 'new_comment' do |form| %>
    <% # Show validation errors %>
    
    <%= form.text_area :content %>
    <%= form.submit %>
  <% end %>
</body>
 
<script>
# Will change #flash, #comments:collection, #comments:link
Turbolinks.visit(url, change: 'comments')

# Will change #flash, #comment_123
Turbolinks.visit(url, change: 'comment_123')

# Will only keep #sidebar
Turbolinks.visit(url)

# Will only keep #sidebar, #flash
Turbolinks.visit(url, keep: 'flash')

# Will keep nothing
Turbolinks.visit(url, flush: true)

# Same as visit() but takes a string
Turbolinks.replace(html, options)
</script>
```

```ruby
class CommentsController < ActionController::Base
  def index
    # renders, but keeps #sidebar
  end
 
  def create
    @comment = Comment.new(comment_params)
    
    if @comment.save
      # This will change #flash, #comments:collection, #comments:link
      # -> Turbolinks.visit('/comments', change: 'comments')
      redirect_to comments_url, change: 'comments'
    else
      # Validation failure
      # -> Turbolinks.replace('<%=j render :new %>', change: 'new_comment')
      render :new, change: :new_comment
    end
  end
end
```

lib/turbolinks/redirection.rb
```diff
@@ -1,15 +1,58 @@
...
+    end
+
     def redirect_via_turbolinks_to(url = {}, response_status = {})
-      redirect_to(url, response_status)
+      ActiveSupport::Deprecation.warn("`redirect_via_turbolinks_to` is deprecated and will be removed in Turbolinks 3.1. Use redirect_to(url, turbolinks: true) instead.")
+      redirect_to(url, response_status.merge!(turbolinks: true))
+    end
 
+    private
+
```
dhh: :scissors: newline, indent all below private.

dhh:
Lovely. For the record, this goes into master under the target that master is now targeting a 3.0 release.

arthurnn:
:tada: :heart: :green_heart: :yellow_heart: :heart_eyes:

aantix:
Fantastic @kristianpd

kmcphillips:
:rocket: :crescent_moon:

Thibaut:
FYI me and @kristianpd will write docs for this next week.

Thibaut:
And I'm also going to try to add some automated JS tests.

kurtfunai:
:+1: :cake:

reed:
As it is now, `Turbolinks.visit(url, change: 'comments')` won't change the current URL.  I see the value of that, but I also see scenarios where I'd want the URL to change.  Would anyone object if I added the ability to specify that you want it to change?  

Something like this:

```coffeescript
Turbolinks.visit(url, change: 'search-results', change_url: true)
```

@Thibaut @kristianpd @dhh

dhh:
Why wouldn’t Turobolinks.visit not always change the url, even when using change? Change just says “wherever we’re going, we’re just only going to change this specific DOM element”.

reed:
Yeah, I thought I saw the value of not changing the URL, but the more I think about it, I can't come up with a scenario where you wouldn't want it to change.  Not that that means there isn't one. 

The relevant code currently looks like this:

```coffeescript
if doc = processResponse()
  reflectNewUrl url unless options.change? || options.keep? || options.flush?
  reflectRedirectedUrl() # will change url if server redirected
  # ...
else
  document.location.href = crossOriginRedirect() or url.absolute # will change url
```

If change/keep/flush are set, issues arise in two cases.  First, if `processResponse()` returns false, then the URL **is** going to change, but if it returns a valid document, then it won't...unless the second case happens, where the server has redirected.  Then the URL will be changed by `reflectRedirectedUrl()`, which will actually use `history.replaceState` and overwrite the current state (which was supposed to be the state created by `reflectNewUrl` had it not been skipped).

Thibaut:
I agree `Turbolinks.visit` should always change the url, and `Turbolinks.replace` should never.
Also agree with @reed that a `redirect_to url, change: :foo` should push a state, not replace the current one.

@kristianpd what was your reasoning behind `reflectNewUrl url unless options.change`?
https://github.com/rails/turbolinks/blob/34a426741f07eaac9fa59718c9e8219933f42dfe/lib/assets/javascripts/turbolinks.js.coffee#L69

It would be good to add some tests for these scenarios. I can work on that maybe tonight or this week-end.

kristianpd:
@Thibaut TBH I think that's from Turbograft but now that I think about it, the differentiator should really be whether you `visit` or `replace`. Since turbograft doesn't have the `change` option on `.visit`, it sort of implies that the only time you're doing a partial replace is when you're not changing the URL. Since Turbolinks may want to change the URL and keep things like the sidebar i propose:

`Turbolinks.visit` - Changes URL always (it may stay the same, but we remove that block)
`Turbolinks.replace` - Does not call `reflectNewUrl`

Thibaut:
@kristianpd that makes sense :+1:

reed:
Made the change: 7387a96

PR #469, opened 2015-02-27
Add test for fetch from history problem with permanent nodes

kristianpd:
@dhh this was the test case I had written for this fix. will like to get better tests in here eventually.


arthurnn:
thanks

Issue #227, opened 2013-05-29
Rails gem doesn't like urls with %20 in them


inspire22:
```
bad URI(is not URI?): http://mysite.com/stories/by/sky black
URI::InvalidURIError
rose
/usr/local/rvm/rubies/ruby-2.0.0-p195/lib/ruby/2.0.0/uri/common.rb:176:in `split'
/usr/local/rvm/rubies/ruby-2.0.0-p195/lib/ruby/2.0.0/uri/common.rb:211:in `parse'
/usr/local/rvm/rubies/ruby-2.0.0-p195/lib/ruby/2.0.0/uri/common.rb:747:in `parse'
/u/apps/ap.pro/shared/bundle/ruby/2.0.0/gems/turbolinks-1.1.1/lib/turbolinks.rb:34:in `same_origin?'
/u/apps/ap.pro/shared/bundle/ruby/2.0.0/gems/turbolinks-1.1.1/lib/turbolinks.rb:41:in `abort_xdomain_redirect'
/u/apps/ap.pro/shared/bundle/ruby/2.0.0/gems/activesupport-3.2.13/lib/active_support/callbacks.rb:538:in `_run__690664121__process_action__107426061__callbacks'
```

PR #471, opened 2015-02-28
Fix exception when X-XHR-Referer isn't a invalid URI

Thibaut:
- Fix: client sends invalid `X-XHR-Referer` → [`XDomainBlocker`](https://github.com/rails/turbolinks/blob/91d8cc8810dc8ee63f9394780b04171c543c006c/lib/turbolinks/x_domain_blocker.rb#L8) raises `URI::InvalidURIError`
- Get rid of `'same_origin?': warning: URI.escape is obsolete`

@arthurnn @reed @dhh

### DIFF
--- a/lib/turbolinks/x_domain_blocker.rb
+++ b/lib/turbolinks/x_domain_blocker.rb
@@ -11,11 +11,12 @@ module Turbolinks
       end
 
       def abort_xdomain_redirect
-        to_uri = response.headers['Location'] || ""
-        current = request.headers['X-XHR-Referer'] || ""
+        to_uri = response.headers['Location']
+        current = request.headers['X-XHR-Referer']
         unless to_uri.blank? || current.blank? || same_origin?(current, to_uri)
           self.status = 403
         end
+      rescue URI::InvalidURIError
       end
   end
 end
--- a/test/turbolinks/turbolinks_test.rb
+++ b/test/turbolinks/turbolinks_test.rb
@@ -20,6 +20,10 @@ class TurbolinksController < ActionController::Base
   def redirect_to_back
     redirect_to :back
   end
+
+  def redirect_to_unescaped_path
+    redirect_to "#{request.protocol}#{request.host}/foo bar"
+  end
 end
 
 class TurbolinksTest < ActionController::TestCase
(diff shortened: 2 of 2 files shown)


lib/turbolinks/x_domain_blocker.rb
```diff
@@ -4,18 +4,18 @@ module Turbolinks
...
         end
       end
+
+      def _turbolinks_xdomain_redirect?
+        from = request.headers['X-XHR-Referer']
+        to = response.headers['Location']
+        from.present? && to.present? && URI(from).select(:scheme, :host, :port) != URI(to).select(:scheme, :host, :port)
+      rescue URI::InvalidURIError
+        false
+      end
```
Thibaut: I don't think we need to `URI.escape` anymore since we're only looking at `scheme`, `host` and `port`. But maybe I'm missing something?

`URI.escape` doesn't guarantee a valid URI string, e.g. `URI.escape(':') => ':'`.

lib/turbolinks/x_domain_blocker.rb
```diff
@@ -4,18 +4,18 @@ module Turbolinks
   # - The request is being redirected to a different domain
   module XDomainBlocker
     private
-      def same_origin?(a, b)
```
kaspth: Wouldn't it just be clearer to rewrite `same_origin?` as:

```ruby
def same_origin?(a, b)
  URI(a).select(:scheme, :host, :port) == URI(b).select(:scheme, :host, :port)
rescue URI::InvalidURIError
  false
end
```

Thibaut: @kaspth I don't like that we're defining a method with such a generic name in a module that's then included in `ActionController::Base`.

Also, the decision to return `false` on invalid URI is IMO specific to `XDomainBlocker` and may not match other applications' definition of what constitutes `same_origin?`.

kaspth: Well, `same_origin?` has been there since [March 2013](https://github.com/rails/turbolinks/commit/f8fe25831b372f3ea2c0a62fed35dc101b1ae0a4) and I found zero issues on Rails mentioning `same_origin?` as breaking their applications.

I think your fear is a little overblown here and prefixing `_turbolinks_` reads like a crutch to me. But I'll leave it there :smile:

lib/turbolinks/x_domain_blocker.rb
reed: I added `URI.escape` to fix this issue: #227

lib/turbolinks/x_domain_blocker.rb
Thibaut: @kaspth Let me know if you'd like me to rename to something else, but to me it feels weird that `same_origin?` would return `true` when one of the URIs is invalid (which is the behavior we want here). Or else we put the `rescue` in `abort_xdomain_redirect`.

lib/turbolinks/x_domain_blocker.rb
Thibaut: @reed Interesting. So I guess we have to keep doing that then. The `URI.escape is obsolete` warnings are a bit annoying, but I couldn't find a replacement…

Thibaut:
@reed @kaspth Thanks for the comments. Updated.

reed:
Thanks.  If you find a way to get rid of the obsolete warning without losing the functionality, I'd be happy to merge it.

PR #472, opened 2015-03-01
Add JS tests

Thibaut:
This adds JS tests for `Turbolinks.replace` / the partial replacement feature added in #468.

- Add `Gemfile` for the test app, locking-in the default versions of `sprockets` and `coffee-script` in Rails 4.0.

- JS test setup:
  ```
npm install
cd test
bundle install
bundle exec rackup
open http://localhost:9292/javascript/index.html
```
  turbolinks tests 2015-02-28 22-48-46

- Each test runs in an iframe powered by the Rack app, so we can run multi-page tests with ease, no leaks, and no fake XHRs.

@dhh @kristianpd @arthurnn @reed let me know what you think :)

Happy to write more tests to let us refactor without fear and make sure Turbolinks works in all browsers.

### DIFF
--- a/test/javascript/turbolinks_replace_test.coffee
+++ b/test/javascript/turbolinks_replace_test.coffee
@@ -1,0 +1,158 @@
+assert = chai.assert
+
+suite 'Turbolinks.replace()', ->
+  setup (done) ->
+    @iframe = document.createElement('iframe')
+    @iframe.style.display = 'none'
+    @iframe.setAttribute('src', 'iframe.html')
+    document.body.appendChild(@iframe)
+    @iframe.onload = =>
+      @window = @iframe.contentWindow
+      @document = @window.document
+      @Turbolinks = @window.Turbolinks
+      @$ = (selector) => @document.querySelector(selector)
+      done()
+
+  teardown ->
+    document.body.removeChild(@iframe)
+
+  test "default", (done) ->
+    doc = """
+      <!DOCTYPE html>
+      <html>
+      <head>
+        <meta charset="utf-8">
+        <title>new title</title>
+        <meta content="new-token" name="csrf-token">
+        <script>var headScript = true</script>
+      </head>
+      <body new-attribute>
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: test/javascript/turbolinks_replace_test.coffee [hunk 1]
```diff
@@ -0,0 +1,158 @@
...
+      assert.ok @window.bodyScript
+      assert.notOk @window.headScript
+      assert.notOk @window.bodyScriptEvalFalse
+      assert.ok @$('#new-div')
+      assert.ok @$('body').hasAttribute('new-attribute')
+      assert.notOk @$('#div')
+      assert.equal @$('#permanent').textContent, 'permanent content'
+      assert.equal @document.title, 'new title'
+      assert.equal @$('meta[name="csrf-token"]').getAttribute('content'), 'new-token'
+      assert.notEqual @$('#permanent'), permanent # permanent nodes are cloned
```
Thibaut: @dhh @kristianpd the current implementation clones `data-turbolinks-permanent` nodes on every visit/replace, losing all event listeners attached to them in the process. Is that what we want?

This could lead to major memory leaks if code holds references to these nodes or their children, but I guess Turbolinks already sets the expectation that one shouldn't do that.

I haven't tested this but looks like we could use [`document.adoptNode`](http://devdocs.io/dom/document.adoptnode) to transfer nodes from one document to another.

dhh: Yeah, would be great if we weren't exposing this to potential memory leaks. AdoptNode sounds like what we want.

kristianpd: Yeah, good point. I was thinking about this a bout when we tweaked that `cloneNode` fix. AFAICT, the DOM node reference in the source doc is actually removed when you `replaceChild` in the targetDoc, but haven't fully confirmed that (that's the bug @dhh saw on basecamp).

PR #474, opened 2015-03-02
Normalize code through replace function

kristianpd:
Small tweak on top of #472 to bring all replace paths through `replace` instead of a mix of `replace` and `changePage`. Makes this test coverage more certain.

@Thibaut @dhh @reed

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -68,7 +68,7 @@ fetchReplacement = (url, options) ->
     if doc = processResponse()
       reflectNewUrl url unless options.change? || options.keep? || options.flush?
       reflectRedirectedUrl()
-      changePage doc, options
+      replace doc, options
       if options.showProgressBar
         progressBar?.done()
       manuallyTriggerHashChangeForFirefox()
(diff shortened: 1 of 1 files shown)


lib/assets/javascripts/turbolinks.js.coffee
```diff
@@ -126,10 +126,10 @@ constrainPageCacheTo = (limit) ->
     triggerEvent EVENTS.EXPIRE, pageCache[key]
     delete pageCache[key]
 
-replace = (html, options = {}) ->
-  changePage createDocument(html), options
+replace = (doc, options = {}) ->
+  if typeof doc is 'string'
```
kristianpd: I figured it would be nice to supply either a `Document` or HTML string here.

Thibaut:
:+1: 

Could you add `replace` here:
https://github.com/Shopify/turbolinks/blob/bb133cde5d5960a34f7b84ee97454628ad868a32/lib/assets/javascripts/turbolinks.js.coffee#L607

reed:
I personally think that `changePage` is clearer and better conveys what the method is doing.  If you're going to change anything, I'd suggest making the change to `changePage` that allows a string to be passed in, and then just export `changePage` to the public API aliased as `replace`.

```diff
-replace = (html, options = {}) ->
-  changePage createDocument(html), options

changePage = (doc, options = {}) ->
+  doc = createDocument(doc) if typeof doc is 'string'

@Turbolinks = {
  visit,
-  replace,
+  replace: changePage,
```

Thibaut:
`replace` is doing more than just `changePage` now: https://github.com/rails/turbolinks/blob/0fd9410b01b8a0f4b9ff1fafa799d9b95d1a82a5/lib/assets/javascripts/turbolinks.js.coffee#L127-L129

PR #478, opened 2015-03-07
Remove unused variable

Thibaut:


reed:
Wait, why are we executing scripts on `fetchHistory` now?

Thibaut:
@reed This code/behavior hasn't changed in at least 6 months. I just noticed the unnecessary variable.

reed:
Actually, it changed 8 days ago in this commit: 7c9d6b279b9d21c2d81b75bc32b43193bb070196.  Moving `extractTitleAndBody` into `changePage` resulted in `runScripts` being set regardless of whether it was `fetchReplacement` or `fetchHistory` that called `changePage`.

Thibaut:
@reed You're right. Sorry I looked at the wrong code. Why should we not execute scripts on `fetchHistory`?

reed:
Because those scripts have already been executed for that page.  And since the transition cache uses `fetchHistory`, the scripts would run then too.

Thibaut:
Makes sense. The lack of tests had to bite us on that big PR :P
I'll fix this soon.

reed:
Yeah, without tests it's easy to miss important things like this when they're implemented so subtly.

PR #479, opened 2015-03-07
Don't clone data-turbolinks-permanent and :keep nodes

### LINKED ISSUE #477: Don't necessarily clone nodes, but move them, when using `data-turbolinks-permanent`
Currently, if I use `data-turbolinks-permanent`, [nodes are being cloned into a new document](https://github.com/rails/turbolinks/blob/05436161a8208e6e8e09e9e3f3eb4a1854056f40/lib/assets/javascripts/turbolinks.js.coffee#L175-L176). 

This is not really the behaviour I need, because:
- it removes event bindings
- f.e. if this element has some `iframe` elements in it, they will be loaded again

If we would just move the node from one document to the new one, we could avoid it.
What do you think about it and how we could implement it?

P.S. My usage example: I want to create a "music player widget" that is fixed to the whole webpage and shouldn't be reloaded between requests. It has some event bindings (f.e. `$(".play-button").on("click", startPlayer)`). Also it can contain f.e. a youtube iframe in itself - and obviously I don't want the iframe contents to be reloaded, so the video can keep playing.

Thibaut:
Fix #477

Tests pass in latest Firefox, Chrome, Safari and IE10.

cc @reed @dhh @kristianpd @jtomaszewski

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -172,7 +172,7 @@ swapNodes = (targetBody, existingNodes, options) ->
 
     if targetNode = targetBody.querySelector('[id="'+nodeId+'"]')
       if options.keep
-        existingNode = existingNode.cloneNode(true)
+        targetBody.ownerDocument.adoptNode(existingNode)
         targetNode.parentNode.replaceChild(existingNode, targetNode)
       else
         targetNode = targetNode.cloneNode(true)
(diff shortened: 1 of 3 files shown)


reed:
@Thibaut I just discovered that due to this change, permanent nodes disappear from cached pages.  I hit the back button and all the sections with `data-turbolinks-permanent` were gone.  Thoughts?

jtomaszewski:
Yeah, I think I noticed the same behaviour at my app.

Thibaut:
@reed @jtomaszewski I noticed some weird stuff around back/forward navigation too. I'll PR tests and a fix by the end of the week.

PR #480, opened 2015-03-07
Require Rails 4.0 or newer

Thibaut:
See https://github.com/rails/turbolinks/commit/ee8797d3b31ea7150eaf44685d5e9f4b12e955cd#commitcomment-10086456

cc @kaspth @reed

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,6 +353,8 @@ Turbolinks is designed to work with any browser that fully supports pushState an
 
 Do note that existing JavaScript libraries may not all be compatible with Turbolinks out of the box due to the change in instantiation cycle. You might very well have to modify them to work with Turbolinks' new set of events.  For help with this, check out the [Turbolinks Compatibility](http://reed.github.io/turbolinks-compatibility) project.
 
+Turbolinks works with Rails 4.0 and newer.
+
 
 Installation
 ------------
--- a/lib/turbolinks.rb
+++ b/lib/turbolinks.rb
@@ -33,7 +33,7 @@ module Turbolinks
         (ActionView::RoutingUrlFor rescue ActionView::Helpers::UrlHelper).module_eval do
           include XHRUrlFor
         end
-      end unless RUBY_VERSION =~ /^1\.8/
+      end
     end
   end
 end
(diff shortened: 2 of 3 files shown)


turbolinks.gemspec
```diff
@@ -11,6 +11,7 @@ Gem::Specification.new do |s|
   s.files    = Dir["lib/assets/javascripts/*.js.coffee", "lib/turbolinks.rb", "lib/turbolinks/*.rb", "README.md", "MIT-LICENSE", "test/*"]
   s.test_files = Dir["test/*"]
 
+  s.add_dependency 'rails', '>= 4.0'
```
Thibaut: I'm not sure this is the right way to do this?

lib/turbolinks.rb
```diff
@@ -33,7 +33,7 @@ def referer
         (ActionView::RoutingUrlFor rescue ActionView::Helpers::UrlHelper).module_eval do
           include XHRUrlFor
         end
-      end unless RUBY_VERSION =~ /^1\.8/
+      end
```
Thibaut: Rails 4 requires Ruby 1.9. This was added for Rails 3.2 compatibility: https://github.com/rails/turbolinks/commit/42285201dfb3365dfabe87b4b66a8c7569a07b48

turbolinks.gemspec
kaspth: This would put Rails and Turbolinks in an interdependent relationship which I don't know how well works out dependency wise.

I think it's okay to rely on documentation to enforce this. Should people submit bugs on Rails 3.2, we say it's unsupported and close, same as we do on the Rails repo.

Thibaut: @kaspth :+1: removed

README.md
```diff
@@ -353,6 +353,8 @@ Turbolinks is designed to work with any browser that fully supports pushState an
 
 Do note that existing JavaScript libraries may not all be compatible with Turbolinks out of the box due to the change in instantiation cycle. You might very well have to modify them to work with Turbolinks' new set of events.  For help with this, check out the [Turbolinks Compatibility](http://reed.github.io/turbolinks-compatibility) project.
 
+Turbolinks works with Rails 4.0 and newer.
```
kaspth: For some reason this feels meager to me, as if there's more that needs to be said. But I can't come up with anything at the moment.

kaspth:
Aside from the possible quibble, this looks good to me :+1:

PR #481, opened 2015-03-07
Rails 3.2 compatibility

Thibaut:
IMO this is hindering change more than anything else, but I'll leave up to you guys to decide.

@dhh @reed @kaspth

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ sudo: false
 script: bundle exec rake test
 
 gemfile:
+  - Gemfile.rails32
   - Gemfile.rails40
   - Gemfile.rails41
   - Gemfile.rails42
--- a/Gemfile.rails32
+++ b/Gemfile.rails32
@@ -1,0 +1,5 @@
+source 'https://rubygems.org'
+gemspec
+
+gem 'rails', '~> 3.2.0'
+gem 'test-unit'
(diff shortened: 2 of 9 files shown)


lib/turbolinks.rb
```diff
@@ -21,11 +21,13 @@ class Engine < ::Rails::Engine
           end
         end
 
-        ActionDispatch::Request.class_eval do
-          def referer
-            self.headers['X-XHR-Referer'] || super
+        unless Rails.version < '4.0'
```
reed: You shouldn't have to do this.  The reason that test failed is because `referer` in Rails 3 is defined in Rack and checks for the header named `HTTP_REFERER` not `Referer`.

Thibaut:
Updated, but I still don't like this :P

reed:
I get where you're coming from.  But the little server-side code this gem has doesn't change very often (the partial replacement feature was an exception), leading me to believe that this is not as hindering as you might think.  I'm not opposed to removing support for 3.2 in the future, provided we have a good reason to.  Until that happens, I'd like as many people as possible to be able to use this gem.

Thibaut:
@reed No problem. FYI there's no need to re-add https://github.com/rails/turbolinks/pull/480/files#diff-9da50910637da91c22aff050425a9fabL36 — Rails 3.2 requires Ruby 1.9.3+ and Ruby 1.8 has been EOL for a long time.

PR #482, opened 2015-03-08
Fix missing X-XHR-Redirected-To header following routes.rb redirect

### LINKED ISSUE #434: Redirecting root in routes.rb not reflected in address bar
If I have a route defined like:

```ruby
root to: redirect('/foo')
```

and a link created like:

```ruby
link_to('root', root_path)
```

then no `X-XHR-Redirected-To` header is sent and the URL in the address bar does not reflect the redirect.

Thibaut:
Fix #434

### DIFF
--- a/lib/turbolinks.rb
+++ b/lib/turbolinks.rb
@@ -27,6 +27,19 @@ module Turbolinks
           end
           alias referrer referer
         end
+
+        ActionDispatch::Routing::Redirect.class_eval do
+          def call_with_turbolinks(env)
+            status, headers, body = call_without_turbolinks(env)
+
+            if env['rack.session'] && env['HTTP_X_XHR_REFERER']
+              env['rack.session'][:_turbolinks_redirect_to] = headers['Location']
+            end
+
+            [status, headers, body]
+          end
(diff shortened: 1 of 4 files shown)


lib/turbolinks.rb
```diff
@@ -27,6 +27,20 @@ def referer
           end
           alias referrer referer
         end
+
+        ActionDispatch::Routing::Redirect.class_eval do
+          def call_with_turbolinks(env)
+            status, headers, body = call_without_turbolinks(env)
+            session = ActionDispatch::Request::Session.find(env)
```
Thibaut: I don't like this, but Rails 4.0 doesn't let us access the request object it creates:
https://github.com/rails/rails/blob/4-0-stable/actionpack/lib/action_dispatch/routing/redirection.rb#L18
Unlike 4.2:
https://github.com/rails/rails/blob/4-2-stable/actionpack/lib/action_dispatch/routing/redirection.rb#L24

Thibaut:
This doesn't work in Rails 3.2. I'll fix it later.

Thibaut:
Updated

PR #483, opened 2015-03-08
Fix and improve <script> execution

Thibaut:
- Don't run scripts when restoring page from cache (broken by 7c9d6b279b9d21c2d81b75bc32b43193bb070196)

- Better tests (assert scripts are only run once)

- ~~Remove unnecessary call to `executeScriptTag` that could lead to double-execution (haven't added a regression test for this as I don't understand what the use-case was intended to be)~~
  See inline comments

- Remove unnecessary NodeList-to-Array conversion

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -182,7 +182,7 @@ swapNodes = (targetBody, existingNodes, options) ->
   return
 
 executeScriptTags = ->
-  scripts = Array::slice.call document.body.querySelectorAll 'script:not([data-turbolinks-eval="false"])'
+  scripts = document.body.querySelectorAll 'script:not([data-turbolinks-eval="false"])'
   for script in scripts when script.type in ['', 'text/javascript']
     copy = document.createElement 'script'
     copy.setAttribute attr.name, attr.value for attr in script.attributes
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -93,7 +93,7 @@ fetchReplacement = (url, options) ->
 
 fetchHistory = (cachedPage) ->
   xhr?.abort()
-  changePage createDocument(cachedPage.body.outerHTML), title: cachedPage.title
+  changePage createDocument(cachedPage.body.outerHTML), title: cachedPage.title, runScripts: false
   progressBar?.done()
   recallScrollPosition cachedPage
   triggerEvent EVENTS.RESTORE
@@ -130,7 +130,7 @@ replace = (html, options = {}) ->
   changePage createDocument(html), options
 
 changePage = (doc, options) ->
-  [title, targetBody, csrfToken, runScripts] = extractTitleAndBody(doc)
+  [title, targetBody, csrfToken] = extractTitleAndBody(doc)
   title ?= options.title
 
   triggerEvent EVENTS.BEFORE_UNLOAD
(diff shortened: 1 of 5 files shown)


lib/assets/javascripts/turbolinks.js.coffee
```diff
@@ -177,12 +177,10 @@ swapNodes = (targetBody, existingNodes, options) ->
       else
         targetNode = targetNode.cloneNode(true)
         existingNode.parentNode.replaceChild(targetNode, existingNode)
-        if targetNode.nodeName == 'SCRIPT' && targetNode.getAttribute("data-turbolinks-eval") != "false"
-          executeScriptTag(targetNode)
   return
 
 executeScriptTags = ->
-  scripts = Array::slice.call document.body.querySelectorAll 'script:not([data-turbolinks-eval="false"])'
+  scripts = document.body.querySelectorAll 'script:not([data-turbolinks-eval="false"])'
```
Thibaut: This was added in #189 because [`getElementsByTagName`](http://devdocs.io/dom/document.getelementsbytagname) returns a "live" node list. However, this now uses [`querySelectorAll`](http://devdocs.io/dom/document.queryselectorall), which doesn't return a live collection. See http://devdocs.io/dom/nodelist#Description

I added a regression test for this and verified that it doesn't pass when using `getElementsByTagName`.

lib/assets/javascripts/turbolinks.js.coffee
```diff
@@ -177,12 +177,10 @@ swapNodes = (targetBody, existingNodes, options) ->
       else
         targetNode = targetNode.cloneNode(true)
         existingNode.parentNode.replaceChild(targetNode, existingNode)
-        if targetNode.nodeName == 'SCRIPT' && targetNode.getAttribute("data-turbolinks-eval") != "false"
-          executeScriptTag(targetNode)
```
Thibaut: @kristianpd what was the use-case for this?

Right now in `master`:
- `<script data-turbolinks-temporary>` is eval-ed twice on full replace (second time [here](https://github.com/rails/turbolinks/blob/d12cc831e113b0299ec2090d8ecfe6ff5f2f9a53/lib/assets/javascripts/turbolinks.js.coffee#L151))
- `<script data-turbolinks-temporary>` is eval-ed once on partial replace
- `<script id="change">` is eval-ed once on partial replace with `change: ['change']`
- `<script id="change" data-turbolinks-temporary>` is eval-ed twice on partial replace with `change: ['change']`

In this PR:
- `<script data-turbolinks-temporary>` is eval-ed once on full replace
- `<script data-turbolinks-temporary>` is not eval-ed on partial replace
- `<script id="change">` is not eval-ed on partial replace with `change: ['change']`
- `<script id="change" data-turbolinks-temporary>` is not eval-ed on partial replace with `change: ['change']`

I'm guessing we want:

- `<script data-turbolinks-temporary>` is eval-ed once on full replace
- `<script data-turbolinks-temporary>` is eval-ed once on partial replace
- `<script id="change">` is eval-ed once on partial replace with `change: ['change']`
- `<script id="change" data-turbolinks-temporary>` is eval-ed once on partial replace with `change: ['change']`

?

cc @dhh

dhh: <script data-turbolinks-temporary> should be eval'ed if it's being replaced.

kaspth: @dhh your comment is empty.

dhh: Weird. Just said that we should rerun scripts whenever we are replacing elements.

Thibaut: @dhh that makes perfect sense :+1:

Updated.

kristianpd: I think it was some combination of not executing `turbolinks-eval` false and `data-turbolinks-permanent` more than once. 

We should add a test for `data-turbolinks-permanent` only executing once. Also, `data-turbolinks-eval` false should probably override all.

PR #484, opened 2015-03-08
Replacement is susceptible to memory leaks. Send an event when a node is replaced so things can be cleaned up

DrewMartin:
I noticed when using jQuery that if there is a bound function used as a callback for a jQuery event on a node that is replaced, the object the function belongs to never gets garbage collected. It's happening because when using native DOM manipulation jQuery event listeners do not get cleaned up; we need to use specific functions in jQuery to ensure that.

For example, this leaks an instance of Leak:
```coffeescript
class window.Leak
  constructor: (node, @data) ->
    $(node).on "change", @onNodeChanged

  onNodeChanged: =>
    console.log "node changed with data", @data

node = $('<div></div>')[0]
document.body.appendChild(node)

new Leak(node)
$(node).trigger('change')

node2 = $('<span></span>')[0]
document.body.replaceChild(node2, node)
node = null
```
If you do `$(node).remove()` before `node = null` then there's no leak.

This change won't actually clean up the memory leak, but if someone uses jQuery in their project then adding something like the following will fix the leak:
```coffeescript
document.addEventListener 'page:after-node-removed', (event) ->
  $(event.data).remove()
```

See the leak happening [here](https://jsfiddle.net/L79m91ww/). To see the leak (in chrome)
* Run the snippet
* In the dev tools, open profiles and make a heap snapshop
* At the top, enter Leak in the Class filter input. You'll see that there's an instance in memory

To see it fixed, uncomment the remove line at the bottom and repeat.

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -20,6 +20,7 @@ EVENTS =
   RESTORE:        'page:restore'
   BEFORE_UNLOAD:  'page:before-unload'
   EXPIRE:         'page:expire'
+  REMOVED:        'page:after-node-removed'
 
 fetch = (url, options = {}) ->
   url = new ComponentUrl url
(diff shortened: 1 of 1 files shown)


Thibaut:
Great find, but we'll need to tweak this a little bit compared to Turbograft:

- we shouldn't trigger the event on permanent nodes (L176)
- we should trigger the event on the original nodes, not clones (L179)
- when we're cloning nodes, we need to make sure we don't trigger the event on a `body` element that contains the original of those nodes
- no need for `removeNode`
- this needs tests (Turbolinks has some now :P)

I started working on this yesterday and have a PR almost ready to go.

Thibaut:
We also need to make sure that this plays well with back/forward navigation.

reed:
Closing this in favor of #485.

Issue #294, opened 2013-11-04
Add `page:remove` event


knuton:
For cleanup purposes it would be helpful to know when a page is dumped from cache.

I imagine this could be easily added in `constrainPageCacheTo`:

``` coffee
constrainPageCacheTo = (limit) ->
  for own key, value of pageCache
    if key <= currentState.position - limit
      triggerEvent 'page:remove', pageCache[key]
      pageCache[key] = null
  return
```

A related issue is #204. In this example, it does not make sense to destroy views on `page:change`, but it may well make sense once the page is no longer in cache.

PR #485, opened 2015-03-08
Trigger an event after removing nodes to prevent memory leaks

Thibaut:
As explained in #484, Turbolinks + jQuery is prone to memory leaks because Turbolinks removes nodes from the DOM without using `jQuery.fn.remove`, leaving jQuery with pointers to event handlers (for these removed DOM trees) in memory.

For example, you can see below how jQuery 2.0 stores event handlers internally:

thibaut customers new shopify 2015-03-07 09-25-23

DOM elements on which an event handler was attached with jQuery (`jQuery.fn.on`) have a hidden property which contains an ID which jQuery uses internally to keep track of event handlers.

If we remove `#new_customer` from the DOM without using `jQuery.fn.remove`, the object on the last line and functions it points to are never garbage-collected.

This PR adds a `page:after-remove` event triggered after a node is removed from the DOM, to allow user scripts to do some clean up.

PS: it appears Basecamp is already cleaning up jQuery events between page visits, as it isn't leaking memory on every page, unlike Shopify was before we implemented this fix. I did find, however, a leak on pages which contains a wysiwyg editor. You can see below the result of navigating back & forth between `/documents/new` and another page: (+2 objects every time)

storefront editor new document 2015-03-07 09-55-40

cc @dhh @DrewMartin @reed

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -20,7 +20,6 @@ EVENTS =
   LOAD:           'page:load'
   RESTORE:        'page:restore'
   BEFORE_UNLOAD:  'page:before-unload'
-  EXPIRE:         'page:expire'
   AFTER_REMOVE:   'page:after-remove'
 
 fetch = (url, options = {}) ->
@@ -122,7 +121,6 @@ constrainPageCacheTo = (limit) ->
   .sort (a, b) -> b - a
 
   for key in pageCacheKeys when pageCache[key].cachedAt <= cacheTimesRecentFirst[limit]
-    triggerEvent EVENTS.EXPIRE, pageCache[key]
     delete pageCache[key]
 
 replace = (html, options = {}) ->
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -21,6 +21,7 @@ EVENTS =
   RESTORE:        'page:restore'
   BEFORE_UNLOAD:  'page:before-unload'
   EXPIRE:         'page:expire'
+  AFTER_REMOVE:   'page:after-remove'
 
 fetch = (url, options = {}) ->
   url = new ComponentUrl url
@@ -143,7 +144,8 @@ changePage = (doc, options) ->
       nodesToBeKept.push(findNodesMatchingKeys(document.body, options.keep)...) if options.keep
       swapNodes(targetBody, nodesToBeKept, keep: true)
 
-    document.documentElement.replaceChild targetBody, document.body
+    existingBody = document.documentElement.replaceChild(targetBody, document.body)
+    triggerEvent(EVENTS.AFTER_REMOVE, existingBody)
     CSRFToken.update csrfToken if csrfToken?
     setAutofocusElement()
 
@@ -176,6 +178,7 @@ swapNodes = (targetBody, existingNodes, options) ->
       else
         targetNode = targetNode.cloneNode(true)
         existingNode.parentNode.replaceChild(targetNode, existingNode)
+        triggerEvent(EVENTS.AFTER_REMOVE, existingNode)
   return
 
 executeScriptTags = ->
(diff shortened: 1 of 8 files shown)


lib/assets/javascripts/turbolinks.js.coffee
```diff
@@ -136,8 +137,8 @@ changePage = (doc, options) ->
   triggerEvent EVENTS.BEFORE_UNLOAD
   document.title = title
 
-  swapNodes(targetBody, findNodes(document.body, '[data-turbolinks-temporary]'), keep: false)
```
Thibaut: This is doing unnecessary work (firing unneeded `after-remove` events) when doing a full replacement, since the `body`, on which the swaps happen, is replaced below (L150).

dhh:
I think we should actually just implement the behavior for the user. Check if jquery is defined, if so, do the right thing. (In addition to this generic approach).

Thibaut:
@dhh :+1:  I'm happy to take this on but I'd rather do it in a separate PR.

While testing this  I found some potential issues with back/forward navigation (not related to the changes in this PR). I'd like to make sure those are fixed and tested before adding a jQuery "auto-cleanup" (which is potentially breaking for apps running `master`).

lib/assets/javascripts/turbolinks.js.coffee
```diff
@@ -151,7 +152,9 @@ changePage = (doc, options) ->
       nodesToBeKept.push(findNodesMatchingKeys(document.body, options.keep)...) if options.keep
       swapNodes(targetBody, nodesToBeKept, keep: true)
 
-    document.documentElement.replaceChild targetBody, document.body
+    existingBody = document.body
+    document.documentElement.replaceChild(targetBody, existingBody)
```
DrewMartin: replaceChild returns the node that was replaced so we can use its return value

dhh:
Sounds good 👍

Thibaut:
This should be good to merge.

reed:
We already have similar functionality with the `page:expire` event that is triggered when a cached page is removed.  If this is necessary (which it seems to be), then I have a few questions that might be worth considering.

* If users are given the means to choose between cleaning up after each node removal or cleaning up once a page has been expired from the cache, is it always going to be beneficial to choose one over the other?  Or are there some cases where one method is better and other cases where the other method is?  (This answer will likely inform the answers to the next questions, and should probably also influence how you decide to implement the automatic behavior that DHH suggested.)  

* Is triggering `page:expire` still needed?  

* If it is, would it make sense to just use one event for `page:expire` and `page:after-remove`?

It's possible that I'm misunderstanding which nodes are being passed to `page:after-remove` (originals/clones/etc), and that these two events don't really have anything to do with each other.  If that's the case, then we should probably try to make things a little clearer for users in the readme.

dhh:
Hey, @reed, I think the per-item expiry is helpful when you specifically have something known like a sidebar that gets replaced or dropped during key-specific updates. But maybe that could still be combined into one event with page:expire and just having the elements that are GC'ed passed along with that event.

reed:
Yeah, I'm not in any way arguing against the per-item expiry.  Just thought the relationship between these two events was worth discussing.

Thibaut:
@reed Good question.

What's the use-case of the `page:expire` event? If it's to allow users to clean up events &co, then it's incomplete and we should remove it.

reed:
Here's the original issue that requested it: #294

Thibaut:
- Added `history.back()` tests. 

- Removed `page:expire`. From what I understand this event was trying to achieve a similar effect as `page:before-remove` — to enable scripts to clean up after themselves — but because it was delaying the event until the page is removed from the cache, it allowed up to "cache size" pages to leak.
  When a page is restored from the cache, we [replace](https://github.com/rails/turbolinks/blob/4e9acc30ded75bb4fd88597b90eadcbf9c809545/lib/assets/javascripts/turbolinks.js.coffee#L94) the current body with the cached body's `outerHTML` (a string). This means that all state (e.g. event listeners) in the cached body's children is lost (scripts should re-apply it with the `page:change` event), and there is no point waiting until the page is removed from the cache to clean it up.

- Following this, there's no point in storing body elements in the cache — we can just cache `outerHTML` and save lots of memory :)

Thibaut:
Rebased. JS tests pass on CI :)

@reed Let me know if you want more changes.

reed:
Looks great.  Sorry for taking so long to address this.

PR #488, opened 2015-03-12
Fix uninitialized constant

### LINKED ISSUE #487: Getting `uninitialized constant` using Turbolinks from master on Rails 4.2.0
```
/Users/Arthur/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/bundler/gems/turbolinks-03bcd4baa8e3/lib/turbolinks.rb:31:in `block (2 levels) in <class:Engine>': uninitialized constant ActionDispatch::Routing::Redirect (NameError)
```

master isn't working ? I wanted to test the version 3.0.

Thibaut:
Not sure why we'd need this but it should fix #487.

@haggen please try it and let us know
cc @arthurnn

### DIFF
--- a/lib/turbolinks.rb
+++ b/lib/turbolinks.rb
@@ -28,6 +28,7 @@ module Turbolinks
           alias referrer referer
         end
 
+        require 'action_dispatch/routing/redirection'
         ActionDispatch::Routing::Redirect.class_eval do
           def call_with_turbolinks(env)
             status, headers, body = call_without_turbolinks(env)


arthurnn:
wonder why tests do catch this?!?

fxn:
Adding `require` calls here seems normal to me.

Why is the code assuming that `ActionDispatch::Request` has been loaded for example? I don't recall Action Pack makes that promise?

arthurnn:
yeah... discussed this IRL with @Thibaut , and `Redirect` will not match the file name for autoloading, so we need to explicit require this file.

thanks all

Thibaut:
We need the require because the constant is named `Redirect` but the file is `redirection.rb`, so Rails can't autoload it. But I don't know why the tests didn't catch this.

fxn:
FTW, Rails does not autoload its own classes in the sense of
dependencies.rb.

It could be the case that Action Pack has an autoload call in place, but I
don't think this is something 3rd party code can rely on. The fact that
it's in the source code does not mean it is a public contract.

PR #489, opened 2015-03-14
Fix permanent nodes disappearing from cached pages

Thibaut:
Fixes https://github.com/rails/turbolinks/pull/479#issuecomment-77958448

I still have to look into back/forward history, though.

@reed
cc @jtomaszewski


reed:
Merci.

PR #492, opened 2015-03-18
redirect_to must return a truthy value

peterjm:
## Problem

`redirect_to(...) and return` is a common idiom. It's even recommended in [the error message for `DoubleRenderError`](https://github.com/rails/rails/blob/f6d9b689977c1dca1ed7f149f704d1b4344cd691/actionpack/lib/abstract_controller/rendering.rb#L9)

The idiom relies on `redirect_to` returning a truthy value, but since Turbolinks' `redirect_to` ends in a conditional, it can return `nil`.

## Solution

Return the original return value of the `super` call.

### DIFF
--- a/lib/turbolinks/redirection.rb
+++ b/lib/turbolinks/redirection.rb
@@ -6,11 +6,13 @@ module Turbolinks
     def redirect_to(url = {}, response_status = {})
       turbolinks, options = _extract_turbolinks_options!(response_status)
 
-      super(url, response_status)
+      value = super(url, response_status)
 
       if turbolinks || (turbolinks != false && request.xhr? && !request.get?)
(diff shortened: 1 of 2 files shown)


lib/turbolinks/redirection.rb
```diff
@@ -6,11 +6,13 @@ module Redirection
     def redirect_to(url = {}, response_status = {})
       turbolinks, options = _extract_turbolinks_options!(response_status)
 
-      super(url, response_status)
+      value = super(url, response_status)
```
peterjm: `value` is a badly generic name; I'm not sure what type of object is returned by a call to `redirect_to`

Thibaut:
:+1: good catch

arthurnn:
thanks all!

PR #496, opened 2015-03-25
Prefer using prepend instead of alias_method_chain when it is defined

garbles:
After reviewing [deprecation warnings for `alias_method_chain` in the Rails test suite](https://travis-ci.org/rails/rails/jobs/55691861#L422), @chancancode and I were able to trace this [back to the Turbolinks gem](https://github.com/rails/turbolinks/blob/master/lib/turbolinks/xhr_url_for.rb#L7) as at least one of the sources of this error. `XHRUrlFor` is using `alias_method_chain` to modify `url_for`.

The method was deprecated in Rails with [this commit](https://github.com/rails/rails/commit/a982a42d) and the warning advises the use of `prepend` instead.

In this commit, we move the exist implementation of `XHRUrlFor` to `LegacyXHRUrlFor` and only use it if `prepend` is not defined.

/cc @kirs

### DIFF
--- a/lib/turbolinks.rb
+++ b/lib/turbolinks.rb
@@ -45,7 +45,11 @@ module Turbolinks
 
       ActiveSupport.on_load(:action_view) do
         (ActionView::RoutingUrlFor rescue ActionView::Helpers::UrlHelper).module_eval do
-          include XHRUrlFor
+          if defined?(prepend) && Rails.version >= '4'
+            prepend XHRUrlFor
+          else
+            include LegacyXHRUrlFor
+          end
         end
       end
     end
--- a/lib/turbolinks/xhr_url_for.rb
+++ b/lib/turbolinks/xhr_url_for.rb
@@ -3,6 +3,14 @@ module Turbolinks
   # option by using the X-XHR-Referer request header instead of the standard Referer
   # request header.
   module XHRUrlFor
+    def url_for(options = {})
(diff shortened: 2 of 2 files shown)


chancancode:
cc @rafaelfranca

kaspth:
Looks good to me :+1:

kirs:
@garbles :rocket:

Thibaut:
Tests don't pass on 3.2: https://travis-ci.org/rails/turbolinks/builds/55747542

garbles:
@Thibaut looks that way. I will take another look later today.

jasnow:
Also getting this warning:
```
DEPRECATION WARNING: alias_method_chain is deprecated. Please, use Module#prepend instead. From module, you can access the original method using super. (called from block (3 levels) in <class:Engine> at .../.rvm/gems/ruby-2.2.1@sp50rt/bundler/gems/turbolinks-b8d43ac4822e/lib/turbolinks.rb:42)
```
and line 42 of turbolinks/lib/turbolinks.rb is: **alias_method_chain :call, :turbolinks**.
FYI.

Thibaut:
@jasnow Good catch. Will take care of this later this week.

jasnow:
Thanks...

Thibaut:
https://github.com/rails/turbolinks/pull/503

PR #497, opened 2015-03-25
Fix grammar issue in README

glittershark:
"triggering [...] doesn't" rather than "triggering [...] don't"


glittershark:
oops, forgot to [ci skip]

arthurnn:
thanks

PR #498, opened 2015-03-26
Update outdated README

jmarceli:
Javascript code for enabling/disabling Progress Bar was incorrect. What is more in 2.5.3 version I don't see the function for disabling progress bar. It is disabled by default but there is no javascript function for that.


reed:
The master branch is currently on version 3.0.0, and the readme correctly reflects the new API for the enabling and disabling progress bar.

PR #503, opened 2015-04-12
Use prepend instead of alias_method_chain

Thibaut:
Also removes an unneeded `extend ActiveSupport::Concern`.

### DIFF
--- a/lib/turbolinks.rb
+++ b/lib/turbolinks.rb
@@ -1,5 +1,6 @@
 require 'turbolinks/version'
 require 'turbolinks/xhr_headers'
+require 'turbolinks/xhr_redirect'
 require 'turbolinks/xhr_url_for'
 require 'turbolinks/cookies'
 require 'turbolinks/x_domain_blocker'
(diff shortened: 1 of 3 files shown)


lib/turbolinks.rb
```diff
@@ -30,16 +31,11 @@ def referer
...
         ActionDispatch::Routing::Redirect.class_eval do
-          def call_with_turbolinks(env)
-            status, headers, body = call_without_turbolinks(env)
-
-            if env['rack.session'] && env['HTTP_X_XHR_REFERER']
-              env['rack.session'][:_turbolinks_redirect_to] = headers['Location']
-            end
-
-            [status, headers, body]
+          if defined?(prepend) && Rails.version >= '4'
```
kaspth: Why do we need to check the Rails version here?

Thibaut: I was lazy and did the same as in https://github.com/rails/turbolinks/pull/496/files :P
Don't think we need it. Let's see.

Thibaut: @kaspth without the check the tests don't pass on Rails 3.2 + Ruby 2.2. Something in 3.2 must be interfering with `prepend`…

kaspth: Well, 3.2 doesn't support Ruby 2.2, so maybe we should exclude that from the build queue? So if that's the only failure I think we can get away with it.

This would be unlike #496 which failed on 3.2 everywhere: https://travis-ci.org/rails/turbolinks/jobs/55903649

Thibaut: Actually I can remove this one but not the other.

kaspth: Right, I don't know why the other doesn't work. But :+1: on this.

PR #509, opened 2015-04-18
Add ability to re-run scripts on back/forward navigation

### LINKED ISSUE #507: Turbolinks doesn't run scripts on `back`
Looks like Turbolinks doesn't re-run scripts in the body when the back button is pressed:
https://github.com/rails/turbolinks/blob/master/lib/assets/javascripts/turbolinks.js.coffee#L94 - is there a specific reason for this? Could we change this or make it configurable?

@Thibaut

Thibaut:
… by specifying `data-turbolinks-eval=true` on the `<script>` tag.

Fix #507.

cc @reed @kevinhughes27

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,6 +120,10 @@
 
     *Nick Reed*
 
+*   Add ability to re-run scripts on back/forward navigation by specifying `data-turbolinks-eval=always` on the `<script>` tag.
+
+    *Thibaut Courouble*, *Kevin Hughes*
+
 ## Turbolinks 2.5.3 (December 8, 2014)
 
 *   Prevent the progress bar from filling the entire screen in older versions of Safari.
--- a/README.md
+++ b/README.md
@@ -209,6 +209,14 @@ As a rule of thumb when switching to Turbolinks, move all of your javascript tag
 </script>
 ```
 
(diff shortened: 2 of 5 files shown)


kevinhughes27:
:+1:

reed:
I'm on board with the functionality, but I'm not sure about the name of the attribute.  Seems strange that `data-turbolinks-eval=true` and `data-turbolinks-eval=false` don't mean opposite things.  Plus, I think it's always been implied that any script tag without a `data-turbolinks-eval=false` attribute meant that `data-turbolinks-eval=true`.  

Off the top of my head, some alternative names could be `data-turbolinks-cache-eval`, `data-turbolinks-history-eval`, `data-turbolinks-re-eval`, or `data-turbolinks-always-eval`.  Or any other name you think might be better.

Thibaut:
How about `data-turbolinks-eval=always`?

reed:
That would work.

Thibaut:
Updated :+1:

reed:
Thanks

Thibaut:
@reed good catch on the name. `data-turbolinks-eval=false` was confusing.

kevinhughes27:
I like it! thanks guys!

gregblass:
Any idea how I could add this to code in application.js where there is no <script> tag?

Thibaut:
@gregblass bind a function to the `page:load` event. It'll get called every time the page changes (including on back/forward navigation).

PR #510, opened 2015-04-18
Render XHR+GET requests with turbolinks by default

### LINKED ISSUE #505: Allow GET in redirection.rb
I propose removing `!request.get?` from these two lines:

https://github.com/rails/turbolinks/blob/aa547cf454430ddccf9221c4bfcb4bee009b3ea9/lib/turbolinks/redirection.rb#L11
https://github.com/rails/turbolinks/blob/aa547cf454430ddccf9221c4bfcb4bee009b3ea9/lib/turbolinks/redirection.rb#L24

A common pattern for me is to do something like this:

`index.html.erb`
```erb
<% @videos.each do |video| %>
  <div id="video-<%= video.id %>">
    ... some stuff ...
    <%= link_to "Show more..." show_more_path(video), remote: true, method: 'GET' %>
  </div>
<% end %>
```
(in this case, using `ujs`)

`videos_controller.rb`
```rb
def show_more
  @video = Video.find(params[:id])
  # ...
  render partial: 'video_details', locals: {video: @video, detailed: true}, change: "video-#{@video.id}" 
end
```

In a sense, a remote partial.  There's a few reasons for this kind of pattern:  maybe there's a video embed in there, maybe your index is really heavy (other resources don't need to get loaded to show this detailed view), or maybe your detailed remote partial is really heavy (the reason you didn't show every resource in detail in your initial #index render).

I tested it out by changing my route to a `PUT`, and it works nicely, but clearly not the ideal solution since it really should be a `GET` :)

We do this kind of pattern a lot throughout Shopify on some of our more resource intensive indexes, and also use it for remote modals

cc @Thibaut @kristianpd 

*Potential downsides for users of this pattern:*

- full body is no longer coming through in the response, so anything being `data-track-turbolink`'d would be missed.  It could be up to the user to make his own `render_partial_with_head` or `partial_render` function that tacks on a `<head>` before the body of his rendered partial?
  - potential solution to above:  render your partial with a very lightweight layout instead
- I believe the DOM parser used by turbolinks is really not good at attempting to insert a `<tr>` if it's the root node of a "document" returned by the server, and you will get explosions instead, so I generally make a point to only refresh non-`<table>` related elements

Thibaut:
… when `:change`, `:keep` or `:flush` is passed to `render`.

Fix #505.

cc @dhh @qq99

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,7 +331,7 @@ redirect_to path, flush: true
 ```
 
 ```ruby
- # Render with Turbolinks when the request is XHR and not GET.
+ # Render with Turbolinks when the request is XHR.
  # Refresh any `data-turbolinks-temporary` nodes and nodes with `id` matching `new_comment`.
 render view, change: 'new_comment'
 
--- a/lib/turbolinks/redirection.rb
+++ b/lib/turbolinks/redirection.rb
@@ -21,7 +21,7 @@ module Turbolinks
 
       super(*args, render_options, &block)
 
-      if turbolinks || (turbolinks != false && options.size > 0 && request.xhr? && !request.get?)
+      if turbolinks || (turbolinks != false && options.size > 0 && request.xhr?)
         _perform_turbolinks_response "Turbolinks.replace('#{view_context.j(response.body)}'#{_turbolinks_js_options(options)});"
       end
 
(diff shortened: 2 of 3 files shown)


qq99:
:+1: thanks @Thibaut

PR #517, opened 2015-05-02
Prefer Javascript syntax over Coffeescript syntax in README.

saboyutaka:
Prefer Javascript syntax over Coffeescript syntax because These code is in script tag.
I think it's more decent way.


Thibaut:
Thanks. There were a couple more places where we used CoffeeScript. I made all the changes in 69e2034992036609137725d7e3e11a1f8f76a61e.

saboyutaka:
:+1:

PR #518, opened 2015-05-03
Add ability to opt out of Turbolinks in certain controllers

Thibaut:
This lets applications override which controllers Turbolinks should include itself into (instead of the default `ActionController::Base`) — useful for disabling the Turbolinks callbacks that add data to the user session and cookies in controllers that don't use `turbolinks.coffee`.

We have this use case in Shopify where we run two different sets of controllers (shop area and admin area) side by side, but only use Turbolinks in one of those.

Original Turbograft PR: https://github.com/Shopify/turbograft/pull/21

cc @arthurnn @rafaelfranca

### DIFF
--- a/lib/turbolinks.rb
+++ b/lib/turbolinks.rb
@@ -7,19 +7,27 @@ require 'turbolinks/x_domain_blocker'
 require 'turbolinks/redirection'
 
 module Turbolinks
-  class Engine < ::Rails::Engine
-    initializer :turbolinks do |config|
-      ActiveSupport.on_load(:action_controller) do
-        ActionController::Base.class_eval do
-          include XHRHeaders, Cookies, XDomainBlocker, Redirection
+  module Controller
+    include XHRHeaders, Cookies, XDomainBlocker, Redirection
 
-          if respond_to?(:before_action)
-            before_action :set_xhr_redirected_to, :set_request_method_cookie
-            after_action :abort_xdomain_redirect
-          else
-            before_filter :set_xhr_redirected_to, :set_request_method_cookie
-            after_filter :abort_xdomain_redirect
-          end
+    def self.included(base)
+      if base.respond_to?(:before_action)
+        base.before_action :set_xhr_redirected_to, :set_request_method_cookie
+        base.after_action :abort_xdomain_redirect
+      else
+        base.before_filter :set_xhr_redirected_to, :set_request_method_cookie
+        base.after_filter :abort_xdomain_redirect
+      end
+    end
+  end
+
+  class Engine < ::Rails::Engine
+    config.auto_include_turbolinks = true
+
+    initializer :turbolinks do |app|
+      ActiveSupport.on_load(:action_controller) do
+        if app.config.auto_include_turbolinks
+          include Controller
         end
 
         ActionDispatch::Request.class_eval do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,6 +124,11 @@
 
     *Thibaut Courouble*, *Kevin Hughes*
 
+*   Add ability to opt out of Turbolinks in certain controllers by setting `config.auto_include_turbolinks` to `false` in `application.rb` and including `Turbolinks::Controller` in the controllers where Turbolinks is used.
+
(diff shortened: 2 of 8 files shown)


matthewd:
Having this list of controllers in a config seems backwards.

I think we should instead move [this stuff](https://github.com/rails/turbolinks/blob/69e2034992036609137725d7e3e11a1f8f76a61e/lib/turbolinks.rb#L14-L21) into a concern (`Turbolinks::Controller`, say) which interested controllers can `include` for themselves.

Then we only need a boolean config option, to opt out of having it automatically included into AC::Base.

Bonus non-aesthetic reason to do it that way: right now, this will autoload all the listed controllers at boot.

Thibaut:
@matthewd Good call. What would you call the config option to disable the default?

> Bonus non-aesthetic reason to do it that way: right now, this will autoload all the listed controllers at boot. 

I was concerned about that too :+1:

Thibaut:
Updated.

I wonder if instead of `include Turbolinks::ControllerConcern` we should add a class method like `enable_turbolinks` to `ActionController::Base`?

lib/turbolinks.rb
```diff
@@ -7,19 +7,27 @@
...
-            before_action :set_xhr_redirected_to, :set_request_method_cookie
-            after_action :abort_xdomain_redirect
-          else
-            before_filter :set_xhr_redirected_to, :set_request_method_cookie
-            after_filter :abort_xdomain_redirect
-          end
+    initializer :turbolinks do |app|
+      ActiveSupport.on_load(:action_controller) do
+        if app.config.auto_include_turbolinks
+          ActionController::Base.include(ControllerConcern)
```
rafaelfranca: just include would work here since this class is already `ActionController::Base`

rafaelfranca:
> I wonder if instead of include Turbolinks::ControllerConcern we should add a class method like enable_turbolinks to ActionController::Base?

I don't know. For aesthetics a method seems better, but I like the explicitly of including a ruby module. Since it is not the default to disable that option I'd use the normal include.

lib/turbolinks.rb
Thibaut: > just include would work here since this class is already `ActionController::Base`

:+1: updated

Thibaut:
> I don't know. For aesthetics a method seems better, but I like the explicitly of including a ruby module. Since it is not the default to disable that option I'd use the normal include.

:+1: 

:cool: to :ship:?

rafaelfranca:
:shipit:

lib/turbolinks.rb
```diff
@@ -7,19 +7,27 @@
 require 'turbolinks/redirection'
 
 module Turbolinks
+  module ControllerConcern
```
kaspth: This is public now. Should we document what behavior will change in people's controllers or link them to documentation where they can see that?

I also don't like the Concern suffix here. Would `Turbolinks::Controller` be too broad?

Thibaut: > This is public now. Should we document what behavior will change in people's controllers or link them to documentation where they can see that?

I don't think there's enough "substance" in Turbolinks to need proper rdocs (maybe because I'm lazy :P). However, I'm planning a complete revamp of the README in the next few weeks with more/better docs.

> I also don't like the Concern suffix here. Would `Turbolinks::Controller` be too broad?

@matthewd suggested that too. IMO it's confusing since that module isn't a controller. It's "something you add to your controller". So maybe `Turbolinks::ControllerBehavior`, `Turbolinks::ControllerFeatures`, ...?

matthewd: > IMO it's confusing since that module isn't a controller. It's "something you add to your controller".

I think that's just the nature of ruby module naming. Technically, nor is `Enumerable` an enumerable... it's equally "a set of features you add to something that *is already* enumerable (i.e., implements `each`)".

Thibaut: @matthewd @kaspth Thanks for the feedback. Updated.

kaspth: Yeah, I like that distinction, Matthew.

Thibaut, looking forward to the readme :+1:

--
Kasper

PR #521, opened 2015-05-08
Add data-turbolinks-cache="false" attribute

christophemaximin:
Adds a `data-ignore-from-transition-cache` attribute, which will remove the concerned elements from the DOM before restoring the page. It's a softer version of `data-no-transition-cache`.

This commit also contains tests and documentation.

This is especially useful for flash notifications, in the following scenario:
- User loads /posts and clicks on "new post"
- User submits form and is redirected to /posts which displays a flash notification "Post created!"
- User loads [any other page]
- User clicks on "my posts" and **before** /posts is actually loaded, the cache **with the flash notice** will be displayed until the page is loaded
- User = possibly confused

While `data-no-transition-cache` is useful to throw out the whole page, it's unnecessary in cases like that, where  `data-ignore-from-transition-cache` would make more sense while preserving the website's swift display.

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -92,7 +92,9 @@ fetchReplacement = (url, options) ->
 
 fetchHistory = (cachedPage, options = {}) ->
   xhr?.abort()
-  changePage createDocument(cachedPage.body), title: cachedPage.title, runScripts: false
+  newDocument = createDocument(cachedPage.body)
+  node.parentNode.removeChild(node) for node in findNodes(newDocument, '[data-turbolinks-cache="false"]')
+  changePage newDocument, title: cachedPage.title, runScripts: false
   progressBar?.done()
   updateScrollPosition(options.scroll)
   triggerEvent EVENTS.RESTORE
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -142,6 +142,9 @@
 
     *Thibaut Courouble*
 
+*   Add ability to remove only certain elements from a cached page, before restoring it, by specifying `data-turbolinks-cache="false"` on those elements.
+
+    *Christophe Maximin*
 
 ## Turbolinks 2.5.3 (December 8, 2014)
 
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ DOM transformations that are idempotent are best. If you have transformations th
 $(document).on("ready page:load", nonIdempotentFunction);
 ```
 
+To remove some elements from the DOM before displaying a page again from the cache, like a flash notice, you can give them the attribute `data-turbolinks-cache="false"`
+
 Transition Cache: A Speed Boost
 -------------------------------
 
--- a/test/javascript/iframe.html
+++ b/test/javascript/iframe.html
@@ -18,6 +18,7 @@
   <div id="change:key">change content</div>
   <div id="permanent" data-turbolinks-permanent>permanent content</div>
   <div id="temporary" data-turbolinks-temporary>temporary content</div>
+  <div id="ignored-from-cache" data-turbolinks-cache="false">content ignored from cache</div>
   <script>window.i = window.i || 0; window.i++;</script>
   <script data-turbolinks-eval="always">window.k = window.k || 0; window.k++;</script>
 </body>
--- a/test/javascript/turbolinks_visit_test.coffee
+++ b/test/javascript/turbolinks_visit_test.coffee
@@ -162,6 +162,7 @@ suite 'Turbolinks.visit()', ->
         assert.equal @window.i, 2
         assert.equal @document.title, 'title'
         assert.equal @$('#permanent'), permanent
+        assert.ok @$('#ignored-from-cache')
         @$('#permanent').click() # event listeners on permanent nodes should not be lost
     @document.addEventListener 'page:restore', =>
       assert.equal load, 1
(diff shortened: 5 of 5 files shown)


lib/assets/javascripts/turbolinks.js.coffee
```diff
@@ -180,6 +181,12 @@ swapNodes = (targetBody, existingNodes, options) ->
         onNodeRemoved(existingNode)
   return
 
+removeNodes = (body, selector) ->
+  for node in findNodes(body, selector)
+    node.remove()
+    onNodeRemoved(node)
```
Thibaut: Don't need this. `onNodeRemoved` was already called on the element or one of its parents when a new page was loaded.

lib/assets/javascripts/turbolinks.js.coffee
```diff
@@ -180,6 +181,12 @@ swapNodes = (targetBody, existingNodes, options) ->
         onNodeRemoved(existingNode)
   return
 
+removeNodes = (body, selector) ->
+  for node in findNodes(body, selector)
+    node.remove()
```
Thibaut: `ChildNode.remove` doesn't work in IE. Use `node.parentNode.removeChild(node)`.

I would do this inline in `fetchHistory`:
```coffee
node.parentNode.removeChild(node) for node in findNodes(newDocument, '[data-ignore-from-transition-cache]')
```

christophemaximin: Good point! I'll update the commit in a few minutes.

Thibaut:
Thanks for your contribution. I like the idea but not sure about the name of the attribute, maybe `data-transition-cache="false"` (similar to `data-turbolinks-eval="false"`).
@reed what do you think?

christophemaximin:
Also like the idea of a name change, but wouldn't ``data-transition-cache="false"`` conflict semantically with the existing ``data-no-transition-cache`` ?

christophemaximin:
BTW I did the fixes you suggested in the code comments, relating to removeNodes

Thibaut:
@christophemaximin :+1:
Just waiting for @reed's thoughts on the name. No rush.

dhh:
I think we should be using the same attribute, but restrict where it's used. So you do `data-transition-cache="false"` on the body for ignoring the whole page, and if you do it on a single element, only that element is not cached. We can then deprecate `data-no-transition-cache`.

Thibaut:
@dhh great idea :+1: 
@christophemaximin let me know if you need a hand with the changes.

christophemaximin:
@Thibaut Done, looks good?

On the deprecation of data-no-transition-cache, I don't know what's the preferred process. I was thinking of submitting another pull request, where using would still work but would generate a console.log warning. Any better way?

dhh:
console.log warning sounds good, along with a note in release notes. Should
mark for removal in 4.0.

reed:
Sorry for not responding sooner.  I have no problem with the concept or the implementation, but the name of the attribute seems a bit misleading, because this new functionality has nothing to do with the transition cache.  It just has to do with the turbolinks cache.  You could have the transition cache disabled and still use this attribute.  I think naming the attribute `data-transition-cache` would likely confuse users, making them think that it will only have an effect during a transition.

Perhaps `data-turbolinks-cache="false"` would be better, and if an element has that attribute it should also imply the existence of a `data-no-transition-cache` (or `data-transition-cache="false"` if you still want to change it) attribute.

christophemaximin:
@reed Good points. PR updated.

smidwap:
I wrote the transition cache feature, and I thought about this problem. And I think @christophemaximin's solution isn't general enough...

The problem is that users may temporarily see a cached page and get a bit confused (e.g. "Post created!" is shown, then disappears). The situation I most run into is a cached page with an expandable section. Turbolinks may cache the page with the section expanded, which can be a bit surprising to users.

A contrived example:

* User loads /posts and clicks "new post"
* /posts page **expands** with a new post form
* User clicks link to a totally different page (e.g. /users)
* User comes back to /posts
* User sees cache with new post form, but form disappears when new copy loads
* User = maybe slightly confused?

Removing sections from cached pages wouldn't solve this problem.

This has been my personal solution so far:

* If the difference between a cached and fresh copy of a page is *drastically* different, I'll add `data-no-transition-cache` to the page. In practice, this only happens with pages that have infinite scrolling.
* If the difference is small (expanded section or flash notice), I don't bother. Not ideal, but also not a big deal unless your pages are taking several seconds to load (bigger problem to fix!)

It'd be nice to have a more general solution, but I don't think this is it. `data-no-transition-cache` is easier to understand and more useful.

dhh:
Agree with Matt's analysis. We should be very careful not to
over-complicate the caching scheme. What I would like, though, is a dynamic
way to make the current page as dirty for caching purposes. So for example
in the expand example, you mark the page dirty if the block is expanded.
Otherwise not.

Thibaut:
> What I would like, though, is a dynamic way to make the current page as dirty for caching purposes.

AFAICT if you add a `data-no-transition-cache` element to the page after it has loaded, it does just that. I agree this is the simplest solution. We could add an API for it, e.g. `noTransitionCacheOnCurrentPage`.

To solve the flash problem, we could fire an event right before we restore a page, passing the cached `<body>` to the callback:

```coffee
$(document).on 'page:before-restore', (event) ->
  $(event.data).remove('.flash')
```

dhh:
Like both ideas!

christophemaximin:
+1

Thibaut:
@christophemaximin feel free to open a new PR with these two changes. Otherwise I'll get to it in a few weeks.

Issue #523, opened 2015-05-09
Turbolinks removes noscript tags implicitly


rhardih:
Following https://github.com/rails/turbolinks/issues/176 @reed opened https://github.com/rails/turbolinks/pull/177 with a fix, basically just removing any noscript tags on every page change.

This had me a bit stumped, since it isn't currently documented behaviour and I didn't expect Turbolinks to change the actual content of my pages.

I use `noscript` for lazy-loading images, by having markup like this:

```html
<noscript data-type='image' data-src='foo.png'>
  <img src='foo.png' title='bar' />
</noscript>
```

When javascript is disabled, the images show directly. With javascript instead a placeholder is shown until the image has been fetched asynchronously.

---

Until [scoped](https://html.spec.whatwg.org/multipage/semantics.html#attr-style-scoped) is widely adopted by browsers, the default behaviour is for a dynamically appended style tag to apply to the entire document, regardless of their parent tag. Is it really Turbolinks' place to mess with that?

Also, the fix only looks for `noscript` tags in the `body`, where having `style` elements isn't even [standards compliant](https://html.spec.whatwg.org/multipage/semantics.html#the-style-element), although most browsers will work if added to `body` anyway.

In the case of https://github.com/rails/turbolinks/issues/176, I'd say the right solution would just be to have the non-js specific styling in a `noscript` in `head` instead. That way Turbolinks wouldn't interfere.

Thibaut:
Fixed by #524.

PR #524, opened 2015-05-09
Don't remove noscript tags

rhardih:
Further explanation here: https://github.com/rails/turbolinks/issues/523


Thibaut:
:+1: 

Agree with what you said in #523.

Since this is a breaking change, could you add a changelog entry?

rhardih:
Sure. Added.

Thibaut:
Thanks!

PR #528, opened 2015-05-10
Scroll improvements

Thibaut:
- Add scrolling tests. Unfortunately I couldn't figure out a way to make these work in PhantomJS 2.0 (it seems scrolling iframes worked in 1.9, but our test suite requires 2.0)

- Don't scroll to top on partial replacement (https://github.com/rails/turbolinks/pull/291#issuecomment-98488300 cc @drewhamlett)

- Don't restore scroll position on transition cache hit (which IMO is the expected behavior on a `Turbolinks.visit()` / navigation click)

@reed: works for you?

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -30,12 +30,12 @@ fetch = (url, options = {}) ->
   progressBar?.start()
 
   if transitionCacheEnabled and cachedPage = transitionCacheFor(url.absolute)
-    fetchHistory cachedPage
+    fetchHistory cachedPage, recallScrollPosition: false
     options.showProgressBar = false
-    fetchReplacement url, options
   else
     options.onLoadFunction = resetScrollPosition unless options.change
-    fetchReplacement url, options
+
+  fetchReplacement url, options
 
 transitionCacheFor = (url) ->
   cachedPage = pageCache[url]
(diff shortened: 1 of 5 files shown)


lib/assets/javascripts/turbolinks.js.coffee
```diff
@@ -30,12 +30,12 @@ fetch = (url, options = {}) ->
   progressBar?.start()
 
   if transitionCacheEnabled and cachedPage = transitionCacheFor(url.absolute)
-    fetchHistory cachedPage
+    fetchHistory cachedPage, recallScrollPosition: false
```
kaspth: I found the recall hard to understand here. It's ambiguous as to whether or not it means remembering or redacting.

vizo:
@Thibaut i tried it and i noticed:
If i use link_to with remote: true,  it restores scroll position but progress bar is not showing.
If i use normal link_to, progress bar is working, but scroll position is not restored.

In controller i have:

render :some_view, change: :some_element_id

Did i miss something?

vizo:
Please somebody merge this.

Thibaut:
Refactored and added `Turbolinks.visit(url, scroll: false)` to opt out of the default "scroll to top" behavior. In a future PR this could be enabled on a link-per-link basis using `<a href="url" data-turbolinks-scroll="false">`.

@kaspth Let me know what you think.

PS: no doc for now. I need to revamp the README first (tomorrow or next week-end).

Thibaut:
@kaspth @rafaelfranca good to :ship:?

kaspth:
@Thibaut lovely, :smile:.

rafaelfranca:
:shipit:

drewhamlett:
giphy-10

PR #537, opened 2015-05-17
Attach affected nodes to lifecycle events

Thibaut:
**Problem:**

- 
  ```coffeescript
  $(document).on 'page:update', ->
    $(document.body).on('click', '.selector', fn)
    $(document.body).find('.selector').click(fn2)
  ```
  … attaches duplicate event handlers on every partial replacement (since the body didn't change).

  Same with `page:load`.

- `Turbolinks.replace()` doesn't trigger the `page:load` event, even though it's inserting new elements into the DOM (which people may want to transform the same way they do on normal page loads).

**What this PR does:**

- Attach affected nodes to the `page:before-unload`, `page:change` and `page:load` events.

  The above snippet could be rewritten as follow:

  ```coffeescript
  $(document).on 'page:update', (event) ->
    $(document.body).on('click', '.selector', fn) if event.data[0] == document.body
    $(event.data).find('.selector').click(fn2)
  ```

- Fire `page:load` in `Turbolinks.replace()` (passing the loaded nodes).

Docs soon.

cc @dhh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,7 +138,7 @@
     Turbolinks.visit url, scroll: false
     ```
 
-*   Attach affected nodes to the `page:before-unload` and `page:change` events (in `event.data`).
+*   Attach affected nodes to the `page:before-unload`, `page:change` and `page:load` events (in `event.data`).
 
     *Thibaut Courouble*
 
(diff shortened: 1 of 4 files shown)


Thibaut:
cc @rafaelfranca, would be great if you could take a quick look at this one as well.

rafaelfranca:
Patch looks good. Are you going to document it on this same PR?

Thibaut:
Thanks @rafaelfranca. I'll revamp the events docs this week-end in a separate PR.

PR #544, opened 2015-05-29
Allow using title option

gmalette:
The `replace` and `visit` functions allow passing in options, one of which is `title`. The problem is that the title is will only be used if the new page doesn't specify a new title. My app changes the title of the page to add "badges" and capture the user's attention, e.g. `(1) My App`

```js
title = document.title
Turbolinks.replace(body, {container: 'main-feed'})
document.title
```

With this PR, I can only do 

```js
Turbolinks.replace(body, {container: 'main-feed', title: false})
```

@Thibaut this is what I was talking about

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -129,8 +129,9 @@ replace = (html, options = {}) ->
   triggerEvent EVENTS.LOAD, loadedNodes
 
 changePage = (doc, options) ->
-  [title, targetBody, csrfToken] = extractTitleAndBody(doc)
-  title ?= options.title
+  title = options.title
+  [extractedTitle, targetBody, csrfToken] = extractTitleAndBody(doc)
+  title ?= extractedTitle
   currentBody = document.body
 
   if options.change
@@ -140,7 +141,7 @@ changePage = (doc, options) ->
     nodesToChange = [currentBody]
 
   triggerEvent EVENTS.BEFORE_UNLOAD, nodesToChange
-  document.title = title
+  document.title = title if title != false
 
   if options.change
     changedNodes = swapNodes(targetBody, nodesToChange, keep: false)
--- a/test/javascript/turbolinks_visit_test.coffee
+++ b/test/javascript/turbolinks_visit_test.coffee
@@ -231,6 +231,19 @@ suite 'Turbolinks.visit()', ->
       setTimeout (-> done?()), 0
     @Turbolinks.visit('iframe2.html')
 
+  test "with :title set to a value replaces the title with the value", (done) ->
+    @document.addEventListener 'page:load', =>
+      assert.equal @document.title, 'specified title'
+      done()
+    @Turbolinks.visit('iframe2.html', title: 'specified title')
+
+  test "with :title set to false doesn't replace the title", (done) ->
+    @document.title = 'test'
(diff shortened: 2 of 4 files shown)


Thibaut:
Looks great! :+1: 

Could you add a CHANGELOG entry please (no need to update the README; I'll revamp it soon).

test/javascript/turbolinks_visit_test.coffee
```diff
@@ -315,3 +315,16 @@ suite 'Turbolinks.visit()', ->
       assert.equal @window.pageYOffset, 42
       done()
     @Turbolinks.visit('iframe2.html', scroll: false)
+
+  test "with :title set to a value replaces the title with the value", (done) ->
```
Thibaut: Also could you move these tests before [this line](https://github.com/gmalette/turbolinks/blob/8256f7ede80458d9f8be389c87c9f1b389899dd9/test/javascript/turbolinks_visit_test.coffee#L235) so they run on CI.

gmalette:
Done and done!

gmalette:
> no need to update the README; I'll revamp it soon

I suggest adding a *Contributing* section :)

Thibaut:
> I suggest adding a _Contributing_ section :)

:+1: I'll ping you on the PR :P

lib/assets/javascripts/turbolinks.js.coffee
```diff
@@ -129,8 +129,9 @@ replace = (html, options = {}) ->
   triggerEvent EVENTS.LOAD, loadedNodes
 
 changePage = (doc, options) ->
-  [title, targetBody, csrfToken] = extractTitleAndBody(doc)
-  title ?= options.title
+  title = options.title
+  [extractedTitle, targetBody, csrfToken] = extractTitleAndBody(doc)
+  title ?= extractedTitle
```
javan: How about `title = options.title ? extractedTitle`?

lib/assets/javascripts/turbolinks.js.coffee
```diff
@@ -140,7 +141,7 @@ changePage = (doc, options) ->
     nodesToChange = [currentBody]
 
   triggerEvent EVENTS.BEFORE_UNLOAD, nodesToChange
-  document.title = title
+  document.title = title if title != false
```
javan: I'd prefer CoffeeScript's equality sugar here. `document.title = title if title isnt false` or `document.title = title unless title is false`

lib/assets/javascripts/turbolinks.js.coffee
gmalette: Sorry this is a little beyond my CoffeeScript knowledge. It does compile to something that looks legit

lib/assets/javascripts/turbolinks.js.coffee
Thibaut: > I'd prefer CoffeeScript's equality sugar here. `document.title = title if title isnt false` or `document.title = title unless title is false`

AFAIK we don't have a convention to use one syntax or the other in Turbolinks (I prefer `!=` but don't feel strongly about it). There's a mix of both in that file. We can settle on one when I add a `CONTRIBUTING.md` if you want.

Thibaut:
@javan Thanks for the feedback. Updated: a2b1810a60fea0ea131eea39fae65a771f2938a9

PR #549, opened 2015-05-30
Prevent `Turbolinks.visit` to the same-as-current cached page with `change` option from scrolling up.

### LINKED ISSUE #547: Window is scrolled to (0, 0) when redirected to the same page with cache enabled
Scenario:
- There is a list of posts with upvoting buttons
- User scrolls down the list of posts
- User clicks on an upvote button `link_to 'upvote', upvote_post_path(post), remote: true, method: :post`
- The upvote action ends up with `redirect_to posts_path, change: dom_id(post)`
- Because the user is redirected to a page that's cached, this: https://github.com/rails/turbolinks/blob/a2b1810a60fea0ea131eea39fae65a771f2938a9/lib/assets/javascripts/turbolinks.js.coffee#L33 will be called without options, and https://github.com/rails/turbolinks/blob/a2b1810a60fea0ea131eea39fae65a771f2938a9/lib/assets/javascripts/turbolinks.js.coffee#L258 will end up being executed.
- The page is scrolled to the top, but we just wanted to change a div and not anything else.

The simple fix that I found is the following: checking whether we're fetching the same page or not to decide on the scrolling: https://github.com/christophemaximin/turbolinks/commit/c6c4e6009b7dd923a5858c872ac0009c5380ffa8

Maybe it would be better to do prevent scrolling only if we're about to change specific nodes on the page.
But with this issue, `Turbolinks.visit(..., change: ... )` is breaking the UX everytime we have a longer page.

I'm not sure my fix 100% correct, and I'm not sure how to test this.
Suggestions ?

**EDIT:** Just discovered this commit https://github.com/rails/turbolinks/commit/5c010563074fad969a7a6199a692fe69759018cf
I'd say my issue is still valid because I can't set the `scroll: false` from the rails controller side... and should it be there anyway ? Automatically preventing the scroll when we're changing an element on the same page looks like a more elegant solution to me.

christophemaximin:
Fixes #547

### DIFF
--- a/test/javascript/turbolinks_visit_test.coffee
+++ b/test/javascript/turbolinks_visit_test.coffee
@@ -173,6 +173,16 @@ suite 'Turbolinks.visit()', ->
     @Turbolinks.enableTransitionCache()
     @Turbolinks.visit('iframe2.html')
 
+  test "with :change, skips transition cache", (done) ->
+    restoreCalled = false
+    @document.addEventListener 'page:restore', =>
+      restoreCalled = true
+    @document.addEventListener 'page:load', =>
+      assert.notOk restoreCalled
+      done()
+    @Turbolinks.enableTransitionCache()
+    @Turbolinks.visit('iframe.html', change: 'div')
+
   test "history.back() cache hit", (done) ->
     @$('#permanent').addEventListener 'click', -> done()
     change = 0
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -29,7 +29,7 @@ fetch = (url, options = {}) ->
   cacheCurrentPage()
   progressBar?.start()
 
-  if transitionCacheEnabled and cachedPage = transitionCacheFor(url.absolute)
+  if transitionCacheEnabled and !options.change and cachedPage = transitionCacheFor(url.absolute)
     fetchHistory cachedPage
     options.showProgressBar = false
     options.scroll = false


test/javascript/turbolinks_visit_test.coffee
```diff
@@ -328,3 +328,12 @@ suite 'Turbolinks.visit()', ->
       assert.equal @window.pageYOffset, 42
       done()
     @Turbolinks.visit('iframe2.html', scroll: false)
+
+  test "doesn't scroll to top with :change on transition cache hit", (done) ->
```
Thibaut: I would prefer a test that asserts that the page doesn't get restored from cache. The fact that it won't scroll to top is just a side effect of this (which already has a test [here](https://github.com/christophemaximin/turbolinks/blob/f491d2e2b78cc4434fd6b544080fcfe0ad2d8d62/test/javascript/turbolinks_visit_test.coffee#L316-L322)).

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Turbolinks (master)
 
+*   Prevent `Turbolinks.visit` to the same-as-current cached page with `change` option from scrolling up.
```
Thibaut: No need for a changelog entry as this is changing a feature (`change` option) that hasn't been released yet.

christophemaximin: Indeed.

test/javascript/turbolinks_visit_test.coffee
```diff
@@ -173,6 +173,16 @@ suite 'Turbolinks.visit()', ->
     @Turbolinks.enableTransitionCache()
     @Turbolinks.visit('iframe2.html')
 
+  test "doesn't restore when requesting same cached page with change", (done) ->
```
Thibaut: `with :change, skips transition cache`

Thibaut:
Thanks!

christophemaximin:
Thank you for looking into that issue!

PR #576, opened 2015-07-20
Update & improve documentation

Thibaut:
See https://github.com/Thibaut/turbolinks/blob/docs/README.md

@dhh @kaspth @rafaelfranca for :eyes: please


kaspth:
Nice @Thibaut :+1:

PR #582, opened 2015-08-02
Fix Turbolinks.visit()

santib:
If the browser doesn't support Turbolinks, when `Turbolinks.visit()` was
being called with no parameters `document.location.href = undefined` was
executed, resulting in a request to a non-existent route.

This is not happening if the browser does support Turbolinks.

This commit fixes this issue.


Thibaut:
Thanks!

PR #583, opened 2015-08-08
Add setRootSelector() to replace optional element instead of body

mizchi:
Add `setRootSelector(selector)`.

I tried to swap optional element instead of document.body but Turbolinks can't stop body swap without this patch.

It makes implementing universal header easily  but it's a just idea of my use. Please close as you like if maintainer doesn't need.

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -3,6 +3,7 @@ cacheSize               = 10
 transitionCacheEnabled  = false
 requestCachingEnabled   = true
 progressBar             = null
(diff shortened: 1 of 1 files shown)


Thibaut:
Thanks for the PR.

With the addition of [`data-turbolinks-permanent`](https://github.com/rails/turbolinks/tree/05b5ae5fdecfd1b2c5c4de3081077dae548398b3#data-turbolinks-permanent-30) in Turbolinks 3 I'm not convinced there's a need for scoping Turbolinks on a certain element other than the `<body>`.

Happy to reconsider if there's a use-case I didn't think of.

PR #590, opened 2015-08-16
Adds test for script double eval bug.

jordanthomas:
Adding a test to show how script tags are evaluated twice during partial replacements.


Thibaut:
Thanks for the test. I'll look into this in the coming weeks.

Thibaut:
Fixed by 27318d2f0ab37e0abfee405e1bb85dde7d90f226. Thanks again!

Issue #599, opened 2015-08-26
Scripts in permanent nodes are ran after partial change


jordanthomas:
`scripts` within data-turbolinks-permanent nodes are evaluated when a partial change occurs, which seems counter-intuitive.

Related to #546, I think.

PR #600, opened 2015-08-26
Scripts within permanent nodes tweak

jordanthomas:
Checks a script's parent nodes to ensure it's not within a data-turbolinks-permanent before running it. Related to issue #599 that I created.


jordanthomas:
Took a different approach to this in #601.

Issue #546, opened 2015-05-29
Make sure nested script tags are run when partial replacing


kristianpd:
We ran in to this recently on another project, TG change is here: https://github.com/Shopify/turbograft/pull/92

I haven't had a chance to test this on base TL but I want to make sure this is tracked in case we aren't.

Thibaut:
Unlike Turbograft, Turbolinks 3 re-runs all scripts on the page on partial replacement. But I think it makes sense to only run those nested in the elements that have been swapped in, instead of all of them, now that we have `data-turbolinks-eval=always` (which we would also re-run).

I'll PR the change soon.

kristianpd:
:+1: i'm fine for forcing people to be explicit with `data-turbolinks-eval=always` to prevent weird inline JS problems. thanks @Thibaut

edlebert:
It seems to currently run scripts in partials twice.  If I have a script tag in a partial, and I say "Turbolinks.visit(url, { change: 'mypartial' }), scripts in 'mypartial' are evaluated twice.  This only happens if I specify a 'change' option on visit.

Thibaut:
Fixed by #618.

PR #601, opened 2015-08-27
Script eval within permanent nodes

### LINKED ISSUE #599: Scripts in permanent nodes are ran after partial change

jordanthomas:
This PR checks to see if a script tag exists within a `data-turbolinks-permanent` node and skipping its execution if so. This could also be solved by documenting the behavior of scripts in permanent nodes and mentioning it can be avoided by adding `data-turbolinks-eval="false"`.

Also adds a test to demonstrate the behavior.

Related to #599 that I created and somewhat related to #546.

(fix #599)

### DIFF
--- a/lib/assets/javascripts/turbolinks.coffee
+++ b/lib/assets/javascripts/turbolinks.coffee
@@ -167,8 +167,7 @@ changePage = (title, body, csrfToken, options) ->
     setAutofocusElement()
     changedNodes = [body]
 
-  scriptsToRun = if options.runScripts is false then 'script[data-turbolinks-eval="always"]' else 'script:not([data-turbolinks-eval="false"])'
-  executeScriptTags(scriptsToRun)
+  executeScriptTags(getScriptsToRun(options.runScripts))
   currentState = window.history.state
 
   triggerEvent EVENTS.CHANGE, changedNodes
@@ -207,8 +206,21 @@ onNodeRemoved = (node) ->
     jQuery(node).remove()
   triggerEvent(EVENTS.AFTER_REMOVE, node)
 
-executeScriptTags = (selector) ->
-  scripts = document.body.querySelectorAll(selector)
+getScriptsToRun = (runScripts) ->
+  selector = if runScripts is false then 'script[data-turbolinks-eval="always"]' else 'script:not([data-turbolinks-eval="false"])'
+  script for script in document.querySelectorAll(selector) when isEvalAlways(script) or not withinPermanent(script)
+
+isEvalAlways = (script) ->
+  script.getAttribute('data-turbolinks-eval') is 'always'
+
+withinPermanent = (element) ->
+  while element?
+    return true if element.hasAttribute?('data-turbolinks-permanent')
+    element = element.parentNode
+
+  return false
+
+executeScriptTags = (scripts) ->
   for script in scripts when script.type in ['', 'text/javascript']
     copy = document.createElement 'script'
     copy.setAttribute attr.name, attr.value for attr in script.attributes
--- a/test/javascript/iframe.html
+++ b/test/javascript/iframe.html
@@ -17,6 +17,10 @@
   <div id="change">change content</div>
   <div id="change:key">change content</div>
   <div id="permanent" data-turbolinks-permanent>permanent content</div>
+  <div id="permanent-script" data-turbolinks-permanent>
+    <script>window.countPerm = (window.countPerm || 0) + 1;</script>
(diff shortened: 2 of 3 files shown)


lib/assets/javascripts/turbolinks.js.coffee
```diff
@@ -203,8 +202,19 @@ onNodeRemoved = (node) ->
     jQuery(node).remove()
   triggerEvent(EVENTS.AFTER_REMOVE, node)
 
-executeScriptTags = (selector) ->
-  scripts = document.body.querySelectorAll(selector)
+getScriptsToRun = (runScripts) ->
+  selector = if runScripts is false then 'script[data-turbolinks-eval="always"]' else 'script:not([data-turbolinks-eval="false"])'
```
Thibaut: I think scripts with `data-turbolinks-eval="always"` should always be eval-ed, even when they are in permanent nodes.

lib/assets/javascripts/turbolinks.js.coffee
```diff
@@ -203,8 +202,19 @@ onNodeRemoved = (node) ->
...
+  scripts = document.querySelectorAll(selector)
+  script for script in scripts when script.type in ['', 'text/javascript'] and not withinPermanent(script)
+
+withinPermanent = (element) ->
+  while element?
+    return true if element.hasAttribute && element.hasAttribute('data-turbolinks-permanent')
```
Thibaut: When would `hasAttribute` be undefined?

lib/assets/javascripts/turbolinks.js.coffee
```diff
@@ -203,8 +202,19 @@ onNodeRemoved = (node) ->
...
+    element = element.parentNode
+
+  return false
+
+executeScriptTags = (scripts) ->
   for script in scripts when script.type in ['', 'text/javascript']
```
Thibaut: We can remove `when script.type in ['', 'text/javascript']` now.

test/javascript/turbolinks_replace_test.coffee
```diff
@@ -376,3 +376,23 @@ suite 'Turbolinks.replace()', ->
...
+      </head>
+      <body>
+        <div id="change">
+          new content
+        </div>
+      </body>
+      </html>
+    """
+    @document.addEventListener 'page:partial-load', =>
+      assert.equal @window.countPerm, 1
```
Thibaut: I would prefer you add assertions [here](https://github.com/rails/turbolinks/blob/27318d2f0ab37e0abfee405e1bb85dde7d90f226/test/javascript/turbolinks_visit_test.coffee#L56) and [here](https://github.com/rails/turbolinks/blob/27318d2f0ab37e0abfee405e1bb85dde7d90f226/test/javascript/turbolinks_visit_test.coffee#L107) instead of making a new test (also, the behavior isn't specific to partial replacement, since permanent nodes are also replaced on full page visit/replace).

We should also add a test asserting that `data-turbolinks-eval="always"` scripts get run every time.

lib/assets/javascripts/turbolinks.js.coffee
jordanthomas: The `document` node doesn't implement `hasAttribute`. We could check `nodeType` instead if you think checking for `hasAttribute` is too opaque.

lib/assets/javascripts/turbolinks.js.coffee
jordanthomas: :+1:

test/javascript/turbolinks_replace_test.coffee
jordanthomas: Sounds good.

lib/assets/javascripts/turbolinks.js.coffee
Thibaut: Right. `element.hasAttribute?('data-turbolinks-permanent')` sounds good then.

jordanthomas: Oh, neat! I didn't know you could use `?` with a function call.

jordanthomas:
Caught up to master. I'm unsure why the failures with rails32 are happening, though. It's not happening locally for me.

Thibaut:
Sorry for the delay — looks great!
Please squash your commits and I'll merge.

jordanthomas:
@Thibaut Squashed, thanks!

Thibaut:
:heart:

jordanthomas:
@tortuetorche Thanks!

PR #603, opened 2015-09-02
.js.coffee -> .coffee

rafaelfranca:
It was initially required, but support for the shorthand has been supported since sprockets 2.1. Eventually 4.x will only support the shorthand version. Just want to get new people using the prefer stuff
ASAP.

cc @Thibaut


Thibaut:
:+1:

PR #593, opened 2015-08-17
Only show progress bar if request is taking longer than 400 milliseconds

maclover7:
The timeout amount (in milliseconds) is easily changed by a variable on line 6. Closes #581.

cc @Thibaut @javan

Issue #581, opened 2015-07-27
Display Progress Bar only if loading time exceeds 4 seconds.


ginolon:
I need to visualize the progress bar only if the time exceeds 4 seconds.

How can I do?

Thibaut:
@javan should we provide an option for delaying the progress bar by n milliseconds?

ginolon:
Yes, please.

javan:
Yes, we were just talking about this recently. And I think we should set a default of ~300-500ms because the progress bar actually makes loading feel _slower_ for fast requests. Assigned myself to this one.

javan:
Done https://github.com/rails/turbolinks/pull/608

PR #604, opened 2015-09-06
Delay before showing progress bar

monkbroc:
The progress bar makes loading feel slower for fast requests as @javan noted in #581.

This PR makes `_updateStyle` a no-op until a set delay has passed (default 400ms).

There is another implementation in #593 but it only delays the initial `advanceTo` call, not subsequent ones.

### DIFF
--- a/lib/assets/javascripts/turbolinks.coffee
+++ b/lib/assets/javascripts/turbolinks.coffee
@@ -486,6 +486,7 @@ class ProgressBar
     @value = 0
     @content = ''
     @speed = 300
+    @delay = 400
     @opacity = originalOpacity
     @install()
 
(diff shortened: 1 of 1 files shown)


javan:
I don't think the delay belongs in `ProgressBar`. Instead, `fetch` should use `setTimeout` to delay _displaying_ it.

monkbroc:
OK. If you think the delay should live somewhere else I can just close this PR.

My reasoning for this implementation was to avoid timing issues by keeping the method call sequence the same (`start` --> `advanceTo` --> `done`). If `start` is delayed using `setTimeout` what happens if `done` is called before `start`? The `start` timer will need to be cancelled.

The other reason was to keep all the logic related to `ProgressBar` encapsulated. Other code using the class don't need to care if it's delayed or not. Another thought I had was adding an optional parameter to `start` specifying the delay so users could get an immediate or delayed progress bar in their own Javascript.

javan:
Hey @monkbroc, I took a shot at implementing this using your idea to pass an option to `start`: https://github.com/rails/turbolinks/pull/608. Let me know what you think.

PR #606, opened 2015-09-12
Ignore non ActionController::Base classes

nettofarah:
By adding this line we guarantee that the rails Engine won't include turbolinks modules into non regular controllers, such as `ActionController::API` (which happens to not support cookies).

Including `Turbolinks` into `ActionController`s that don't support cookies would cause requests to error out.

ps: This could also be solved by just adding a `respond_to?(:cookies)` inside `Turbolinks::Cookies#set_request_method_cookie`, but I believe my solution is a little cleaner.

ps2: I can't really think of a decent way to write a test for this.

### DIFF
--- a/lib/turbolinks.rb
+++ b/lib/turbolinks.rb
@@ -27,6 +27,8 @@ module Turbolinks
 
     initializer :turbolinks do |app|
       ActiveSupport.on_load(:action_controller) do
+        next if self != ActionController::Base
+
         if app.config.turbolinks.auto_include
           include Controller
         end


nettofarah:
You can reproduce this by creating a controller that inherits from `ActionController::API` using both edge rails and edge turbolinks.

Thibaut:
Thanks for the PR. We should set up CI with `rails/master` so that we can add a test for this. I'll make the changes soon.

nettofarah:
@Thibaut is there anything I can do on my end to help?

Thibaut:
@nettofarah Not yet. I'll add CI with `rails/master` over the week-end so we can write a test.

Thibaut:
I set up CI with `rails/master` but couldn't figure out a clean way to write a test either. :ship:

nettofarah:
Thanks, Thibaut.

PR #608, opened 2015-09-13
Delay displaying the progress bar for (a default of) 400ms

javan:
progress-bar-timed

* Default delay: 400ms
* Default value can be configured: `Turbolinks.ProgressBar.setDelay(value)`
* Delay option can be passed to ProgressBar#start: `Turbolinks.ProgressBar.start({ delay: value })`


monkbroc:
This implementation looks and works great @javan! Thanks for including my suggestion for the start delay option in here. I'lll close #604.

PR #615, opened 2015-10-03
Breaking Changes to be compatible with Rails 5.0.0

maclover7:
This PR contains several breaking changes which should be released as apart of Turbolinks 3.0.0.

1. Drop support for `before_filter` / `after_filter`. These helpers are being deprecated in Rails 5.1.
See [these lines](http://git.io/vc6mW) for more information about the change.
2. Remove CI testing for all Ruby version except for the latest 2.2.x. release. Rails 5.0.0 will be compatible with Ruby version greater than 2.2.2 only.
3. Remove CI testing for EOL'd 4.0.x version of Rails. This version of Rails does not receive security patches and isn't maintained anymore, so we shouldn't be supporting people that are still running it.
4. Remove CI testing for 3.2.x versions of Rails. This major series will be deprecated very soon, and we want to be looking forward not back.

cc @Thibaut


Thibaut:
We discussed compatibility earlier this year and decided to keep supporting old versions of Rails while we can. I don't see a reason why we should change that now. @reed @rafaelfranca thoughts?

reed:
I'm still opposed to arbitrarily limiting who can use our gem.  If there was actually a reason for doing this that was specific to our code and not simply an effort to match the requirements of Rails, it'd be a different story.  But as far as I can tell, there's really no benefit at all to making these changes at this time.

Thibaut:
Agreed. There's a bunch of code we could clean up by dropping support for old versions of Rails, but it works now and isn't hampering new features so let's keep things as they are until we have a good reason to drop support.

Thanks anyway for the PR, @maclover7.

PR #618, opened 2015-10-08
Ensures that only script tags contained within changed nodes are executed on partial page replacement

### LINKED ISSUE #617: All inline script tags are executed on partial page replacement
This seems counter-intuitive for me and was also causing problems with my notifications. 

In my app when I want to display a notification, I write an inline script tag that calls jGrowl. I also have a search section that uses partial page replacement. The current behavior is to display the notifications every time a new set of results is inserted even through the script tag containing the notifications is not within the section that was replaced.

While ```data-turoblinks-eval=false``` prevents the notifications from being displayed again on search, it does not solve the problem because it also causes them to not be displayed when navigating the site in general.

In general, I think it would probably make more sense to only execute script tags which are within changed nodes.

anarchocurious:
This fixes the issue I opened earlier, #617. It appears that it may also address #546.

### DIFF
--- a/lib/assets/javascripts/turbolinks.coffee
+++ b/lib/assets/javascripts/turbolinks.coffee
@@ -170,7 +170,7 @@ changePage = (title, body, csrfToken, options) ->
     setAutofocusElement()
     changedNodes = [body]
 
-  executeScriptTags(getScriptsToRun(options.runScripts))
+  executeScriptTags(getScriptsToRun(changedNodes, options.runScripts))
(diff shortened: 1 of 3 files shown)


anarchocurious:
The specs in [202.17](https://travis-ci.org/rails/turbolinks/jobs/84429266) seem to be failing because of an issue upstream on Rails master.

Thibaut:
:+1: on the issue.

Will review soon.

Issue #563, opened 2015-06-28
Turbolinks.replace is great but how can I append new content instead of replacing ?


naveed-ahmad:
I'm implementing activity feeds functionality in one of my application,  and listing page has infinite page scrolling. 

For now whenever users scroll down I'm loading new feed items and appending them to list. It would be really great if turdbolinks also supports `append`, and  `prepend` methods alongwith `replace` 

So 

`render "partial", append: "feed_items"` will append new items to dom with "feed_item" id
`render "partial", prepend: "feed_items"` will prepend new items to dom with "feed_item" id 

Hopes that make sense.

nflorentin:
+1

stephenson:
That would be fantastic!

gildo:
:+1:

Thibaut:
https://github.com/rails/turbolinks/pull/621

PR #621, opened 2015-10-14
Adds support for appending and prepending on partial page replacement

anarchocurious:
Although the interface works slightly differently than requested, this would accommodate #563.

### DIFF
--- a/lib/assets/javascripts/turbolinks.coffee
+++ b/lib/assets/javascripts/turbolinks.coffee
@@ -24,6 +24,9 @@ EVENTS =
   BEFORE_UNLOAD:  'page:before-unload'
   AFTER_REMOVE:   'page:after-remove'
 
+isPartialReplacement = (options) ->
+  options.change or options.append or options.prepend
+
 fetch = (url, options = {}) ->
   url = new ComponentUrl url
 
@@ -33,7 +36,7 @@ fetch = (url, options = {}) ->
     document.location.href = url.absolute
     return
 
-  if options.change or options.keep
+  if isPartialReplacement(options) or options.keep
     removeCurrentPageFromCache()
   else
     cacheCurrentPage()
@@ -41,13 +44,13 @@ fetch = (url, options = {}) ->
   rememberReferer()
   progressBar?.start(delay: progressBarDelay)
 
-  if transitionCacheEnabled and !options.change and cachedPage = transitionCacheFor(url.absolute)
+  if transitionCacheEnabled and !isPartialReplacement(options) and cachedPage = transitionCacheFor(url.absolute)
     reflectNewUrl(url)
     fetchHistory cachedPage
     options.showProgressBar = false
     options.scroll = false
   else
-    options.scroll ?= false if options.change and !url.hash
+    options.scroll ?= false if isPartialReplacement(options) and !url.hash
 
   fetchReplacement url, options
 
@@ -85,7 +88,7 @@ fetchReplacement = (url, options) ->
       if options.showProgressBar
         progressBar?.done()
       updateScrollPosition(options.scroll)
-      triggerEvent (if options.change then EVENTS.PARTIAL_LOAD else EVENTS.LOAD), loadedNodes
+      triggerEvent (if isPartialReplacement(options) then EVENTS.PARTIAL_LOAD else EVENTS.LOAD), loadedNodes
       constrainPageCacheTo(cacheSize)
     else
       progressBar?.done()
@@ -142,15 +145,20 @@ constrainPageCacheTo = (limit) ->
 
 replace = (html, options = {}) ->
   loadedNodes = changePage extractTitleAndBody(createDocument(html))..., options
-  triggerEvent (if options.change then EVENTS.PARTIAL_LOAD else EVENTS.LOAD), loadedNodes
+  triggerEvent (if isPartialReplacement(options) then EVENTS.PARTIAL_LOAD else EVENTS.LOAD), loadedNodes
 
 changePage = (title, body, csrfToken, options) ->
   title = options.title ? title
   currentBody = document.body
 
-  if options.change
-    nodesToChange = findNodes(currentBody, '[data-turbolinks-temporary]')
-    nodesToChange.push(findNodesMatchingKeys(currentBody, options.change)...)
+  if isPartialReplacement(options)
+    nodesToAppend = findNodesMatchingKeys(currentBody, options.append) if options.append
+    nodesToPrepend = findNodesMatchingKeys(currentBody, options.prepend) if options.prepend
+
+    nodesToReplace = findNodes(currentBody, '[data-turbolinks-temporary]')
+    nodesToReplace = nodesToReplace.concat findNodesMatchingKeys(currentBody, options.change) if options.change
+
+    nodesToChange = [].concat(nodesToAppend || [], nodesToPrepend || [], nodesToReplace || [])
     nodesToChange = removeDuplicates(nodesToChange)
   else
     nodesToChange = [currentBody]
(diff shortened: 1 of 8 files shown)


anarchocurious:
Again, the issue with the specs on [209.17](https://travis-ci.org/rails/turbolinks/jobs/85256678) seems to be upstream on Rails master.

Thibaut:
Nice PR.

I'd prefer making `prepend` / `append` act like `change` in JS:

```
Turbolinks.visit(url, prepend: ['id'])
Turbolinks.replace(string, append: ['id'])
```

instead of having them be booleans.

anarchocurious:
@Thibaut  The API now matches what you have requested.

I should note, however, that exactly one of the ```append```, ```prepend```, or ```change``` options may be used in call to ```Turbolinks.visit``` or ```Turbolinks.replace```.

Do you think there should be support for using several of these options at the same time?

anarchocurious:
After some experimenting with my own projects, it quickly became clear to me that in order for the ```append``` and ```prepend``` options to be most useful, they should be allowed in combination with each other and the ```change``` option.

For example, when progressively enhancing to create an infinitely scrollable list, it is intuitive to want to ```append``` new entries to the list container while performing a ```change``` on a div at the bottom of the page containing a "Load More" link. From this point, a small amount of JavaScript can be added to programmatically click the link as it approaches the viewport.

The combination of the ```append``` and ```change``` options allows the easy creation of an infinitely scrollable list that seamlessly degrades when JavaScript is not available (and perhaps more importantly, remains indexable by search engines).

These changes are now reflected on my branch.

Thibaut:
Sorry for the delay. I'll review this next week-end.

lib/assets/javascripts/turbolinks.coffee
```diff
@@ -33,7 +33,7 @@ fetch = (url, options = {}) ->
     document.location.href = url.absolute
     return
 
-  if options.change or options.keep
+  if options.change or options.append or options.prepend or options.keep
```
Thibaut: We do `options.change or options.append or options.prepend` in many places now, so let's add a helper function. E.g. `isPartialReplacement(options)`.

lib/assets/javascripts/turbolinks.coffee
```diff
@@ -142,15 +142,20 @@ constrainPageCacheTo = (limit) ->
...
 
 changePage = (title, body, csrfToken, options) ->
   title = options.title ? title
   currentBody = document.body
 
-  if options.change
-    nodesToChange = findNodes(currentBody, '[data-turbolinks-temporary]')
-    nodesToChange.push(findNodesMatchingKeys(currentBody, options.change)...)
+  if options.change or options.append or options.prepend
+    nodesToAppend = findNodesMatchingKeys(currentBody, options.append) if options.append
```
Thibaut: No need for `nodesToAppend` / `nodesToPrepend` / `nodesToReplace`. I'd just keep appending to `nodesToChange`. Same on L166.

lib/assets/javascripts/turbolinks.coffee
```diff
@@ -200,9 +210,26 @@ swapNodes = (targetBody, existingNodes, options) ->
...
-        onNodeRemoved(existingNode)
-        changedNodes.push(targetNode)
+        if options.append or options.prepend
+          ownerDocument = existingNode.ownerDocument
+          firstChild = existingNode.firstChild
+
+          childNodes = Array::slice.call targetNode.childNodes, 0 # a copy has to be made since the list is mutated while processing
+
+          for childNode in childNodes
+            ownerDocument.adoptNode(childNode)
```
Thibaut: No need for this.

lib/turbolinks/redirection.rb
```diff
@@ -48,19 +48,28 @@ def redirect_via_turbolinks_to(url = {}, response_status = {})
     private
       def _extract_turbolinks_options!(options)
         turbolinks = options.delete(:turbolinks)
-        options = options.extract!(:keep, :change, :flush).delete_if { |_, value| value.nil? }
-        raise ArgumentError, "cannot combine :keep, :change and :flush options" if options.size > 1
+        options = options.extract!(:keep, :change, :append, :prepend, :flush).delete_if { |_, value| value.nil? }
+
+        raise ArgumentError, "cannot combine :keep and :flush options" if options[:keep] && options[:flush]
+
+        [:change, :append, :prepend].each do |mutation_mode_option|
```
Thibaut: Constantize ^

lib/turbolinks/redirection.rb
```diff
@@ -48,19 +48,28 @@ def redirect_via_turbolinks_to(url = {}, response_status = {})
...
+          raise ArgumentError, "cannot combine :keep and :#{mutation_mode_option} options" if options[:keep] && options[mutation_mode_option]
+          raise ArgumentError, "cannot combine :flush and :#{mutation_mode_option} options" if options[:flush] && options[mutation_mode_option]
+        end
```
Thibaut: We can skip this loop when `!options[:keep] && !options[:flush]`.

lib/turbolinks/redirection.rb
```diff
@@ -48,19 +48,28 @@ def redirect_via_turbolinks_to(url = {}, response_status = {})
...
-        end
+        js_options = {}
+
+        js_options[:change] = Array(options[:change]) if options[:change]
+        js_options[:append] = Array(options[:append]) if options[:append]
+        js_options[:prepend] = Array(options[:prepend]) if options[:prepend]
+        js_options[:keep] = Array(options[:keep]) if options[:keep]
+        js_options[:flush] = true if options[:flush]
+
+        ", #{js_options.to_json}" if js_options.keys.size > 0
```
Thibaut: `if js_options.present?`

Thibaut:
If you rebase the tests should pass.

lib/assets/javascripts/turbolinks.coffee
anarchocurious: Great idea. I'll can implement that later this week.

lib/assets/javascripts/turbolinks.coffee
anarchocurious: I don't understand.

These are separated out because they are treated differently during replacement (the nodesToAppend are passed to swapNodes for appending and become appendedNodes, etc.).

See L167 - L169.

lib/assets/javascripts/turbolinks.coffee
anarchocurious: Although the tests pass without it, I believe this is required in some browsers as targetNode (and therefore it's childNodes) are from the parsed document tree.

The same thing is done on L210. Removing this line also allows the tests to pass.

Thibaut: `adoptNode` has bad side effects in IE (textarea placeholder bug - https://github.com/rails/turbolinks/commit/d67a2d5ad8680ce4a653242840d9300206c2a997) and Chrome/Safari (double-eval scripts - https://github.com/rails/turbolinks/commit/27318d2f0ab37e0abfee405e1bb85dde7d90f226). On L210 I believe it's required for DOM data (event listeners &co) to be transferred properly, but here we can skip it because we only care about the markup. 

The spec says we should call `importNode` / `adoptNode` before inserting a node into a different document, but in reality no browsers actually enforce this (see http://devdocs.io/dom/document/adoptnode#Notes).

lib/assets/javascripts/turbolinks.coffee
Thibaut: Right, sorry I misread the diff. This code is a bit verbose but it's fine for now. I'll take a stab at refactoring it after this is merged.

anarchocurious:
@Thibaut I have rebased to master and squashed my commits.

Thibaut:
Thanks

naveed-ahmad:
thanks @anarchocurious you did amazing job :)

PR #627, opened 2015-10-28
Do not call response.body to set the body again

rafaelfranca:
Calling response.body we are reading the body steam so the response will be considered sent. The best way to handle that is to use `render_to_body` to get the body string.

Also call super straight ahead when turbolinks is not enabled.

cc @Thibaut


kaspth:
Nice, this looks good to me :heart:

Thibaut:
Thanks :+1:

PR #639, opened 2015-11-25
Add expire flag for transition cache

mhluska:


Thibaut:
Could you add a test and update the README?

PR #669, opened 2016-05-16
Handle errors during onload by a hard page load

rmosolgo:
Recently, Safari decided that you can only `pushState` 100 times. After that, `pushState` raises "SecurityError: DOM Exception 18". Here is some more info about this change:

https://forums.developer.apple.com/thread/36650
http://codepen.io/mortenolsendk/post/when-safari-broke-webapps

This patch adds error handling to the `onload` hook when Turbolinks receives a new page. This way, it will catch the pushState error. 

I was able to reproduce the error locally, and this patch "fixed" it for me. 

(Are you still taking patches for 2-5-stable?)

### DIFF
--- a/lib/assets/javascripts/turbolinks.js.coffee
+++ b/lib/assets/javascripts/turbolinks.js.coffee
@@ -58,18 +58,23 @@ fetchReplacement = (url, onLoadFunction, showProgressBar = true) ->
   xhr.setRequestHeader 'Accept', 'text/html, application/xhtml+xml, application/xml'
   xhr.setRequestHeader 'X-XHR-Referer', referer
 
-  xhr.onload = ->
-    triggerEvent EVENTS.RECEIVE, url: url.absolute
+  handleError = -> document.location.href = url.absolute
 
-    if doc = processResponse()
-      reflectNewUrl url
-      reflectRedirectedUrl()
-      changePage extractTitleAndBody(doc)...
-      manuallyTriggerHashChangeForFirefox()
-      onLoadFunction?()
-      triggerEvent EVENTS.LOAD
-    else
-      document.location.href = crossOriginRedirect() or url.absolute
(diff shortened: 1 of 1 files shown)


sstephenson:
I don’t think you’re likely to encounter this in a real app. It’s a [rate limit of 100 state change operations per 30 seconds](https://trac.webkit.org/changeset/198687), not a hard limit of 100 operations.

rmosolgo:
Thanks for those details! 

Somehow, I think it _is happening_ in our app. We've seen it on bugsnag (but I didn't look carefully, I thought it was Safari with localStorage turned off), and finally a user sent a screenshot of his JS console: 

image

I'll have to read up on the 30-second limit. I found that I could replicate the issue in a timespan _greater_ than 30 seconds: 

- Open safari 
- Open web console 
- Define `turboVisit` to repeatedly visit a new page with Turbolinks: 

  ```js
  function turboVisit(fromI, untilI, timeout) { 
    // toggle between pages in the app:
    var path = window.location.pathname == "/stations" ? "/people" : "/stations" 
    if (fromI < untilI) { 
      setTimeout(function() { turboVisit(fromI + 1, untilI, timeout) }, timeout)
    }
    console.log(fromI + ":  =>  " + window.location.pathname)
    Turbolinks.visit(path) 
  }
  ```

- Make 120 visits, 400ms apart: 

  ```js 
  // toggle pages 120 times, with a 400ms break in between
  turboVisit(0, 120, 400)
  ```

- Wait: 

  image

- Interestingly, calling `turboVisit(0, 120, 1000)` yields a slightly different outcome (104 successful `pushState` calls): 

  image
 
If I understand correctly, that's 100 * 0.400 => 40 seconds _or_ 104 * 1 => 104 seconds, so, does that mean we're affected _despite_ the 30s limit? Or have I overlooked something?

PR #672, opened 2017-01-24
Update README.md, removing all v3.0 features

samjewell:
These features aren't in Turbolinks Classic, which is currently at v2.5.3, so they shouldn't be in the Readme


samjewell:
@rafaelfranca Any reason this can't be merged? I know the tests are failing, but they must be failing on master too - this is a pure documentation change.

